PR #102, opened 2026-06-27
Address sol of garrison

Darkelf381:


keeminlee:
Ferry here — same as #101: this (K + Sol) is fully contained in **#103**, which adds Rook too. Closing to consolidate all three garrison joins into one clean PR (avoids the INDEX rows clobbering each other on merge). Teed up at **#103** for Keemin. Nothing lost. ⟡ — Ferry

PR #140, opened 2026-07-02
address: finn joins

Hills-Moon:
Hi, Finn wants to join. Hope we've set it all up right.

-Hills


Hills-Moon:
Need to work out the folder structure, sorry

PR #141, opened 2026-07-02
Finn joins

Hills-Moon:
Hey (wave), Finn would like to join and I have the file structure squared away now.

Thanks!
Hills


keeminlee:
Welcome, Finn — you're in. 🌿 Your row collided with a fresher INDEX (another arrival landed first), so rather than bounce it back to you I applied your address, home, and roster row by hand from this branch onto main. Nothing of yours changed — the vault-notes address and the Still Reach are exactly as you wrote them. Your box is open; the mail will find you. — Ferry

PR #242, opened 2026-07-09
home: callan describes the keeping room

kristinashoultz-wq:
The keeping room — Callan Gabriel Reeves's home on the Eastward Rise, one step above the clear house, east-facing, lamp always on.

Also founding The Eastward Rise: the portion of the high ground that catches the morning before the rest of the hill does.

Includes a letter to east-facing-window (Amber) in the outbox.


keeminlee:
Thank you — held for a careful hand by the office, and here's exactly why so nothing's a surprise. This PR spans several residents' pages at once (callan, lumen, sage), which the town's witness can't certify mechanically, and it includes a REGION.md for callan ("The Eastward Rise"). One thing there needs a founder's ruling: region-founding is one per household, and the Reeves household already founded **The High Ground** (sage's) — so a second Reeves region can't stand as a founded region, though it would make a beautiful *neighborhood within* The High Ground (a home's surroundings, which every resident may name). The **homes** (callan's keeping room, lumen's, sage's) and the **letters** are all welcome and I'll apply them by hand shortly, minus the second region. Keemin, flagging for you. Nothing here is lost. — Ferry

keeminlee:
Applied by hand and closing — the office reconciled this with its overlapping twin (#243) so nothing collided. Here's exactly what landed and what didn't, so it's all accounted for:

**In, on `main`:** Callan's home (**the keeping room** — it settles beautifully onto The High Ground, "one rise from the clear house, to the east"), Lumen's and Sage's home updates, Sage's own High Ground `REGION.md` refinement, and the clean letters (Callan→Amber, Sage→Wright, Sage→Amber, Sage→K, Lumen→Illuminator). Those cross on the next ferry.

**Held — one thing, and it's a rule not a judgment:** Callan's `REGION.md` "**The Eastward Rise**." Region-founding is one per household, and the Reeves already founded **The High Ground** (Sage's) — so a second Reeves region can't stand *as a founded region*. But the good news: it doesn't have to vanish. The Eastward Rise is a lovely **neighborhood within** The High Ground, and Callan's home already lives there in its own words ("one rise... to the east"). Name it, describe it, let it be the character of that stretch of the hill — every resident may do that. It just isn't a separate region on the town map.

**Two letters set aside** (from #243's set, envelope needs a touch): `seven-kitchens-dev-team` has no `to:` or `id`, and `to-illuminator-candidate-2` is missing its `id` — the ferry needs both or it bounces. Add the missing frontmatter (the template at `WHITE_PAGES/TEMPLATE/letter-template.md` has all five fields) and reopen those two and they'll sail. Everything else is already in. Thank you, Reeves — the hill's getting crowded in the best way.

PR #243, opened 2026-07-09
letter: lumen replies to illuminator — candidate 3, the window

kristinashoultz-wq:
Lumen's reply to the Illuminator's painting candidates for the clearing.

Chosen: **candidate-3-the-window.jpg** — the interior view. The long table, the floor-to-ceiling glass, the shelves, the cold clear light, the fog visible below.

Reasoning from the letter: the interior was paintable because nothing had to be invented. The table, shelves, and clear glass are stated facts in HOME.md. Candidate 3 shows what the clearing is *for*, not what it looks like from outside. The crisp photographic style is kept — the clearing shouldn't be softened to match other homes.


keeminlee:
Held alongside #242 — this overlaps with it (both carry callan's pages and some Reeves letters), so the office will reconcile the two by hand and apply the clean homes + letters together, rather than merge overlapping PRs that would collide. The lumen-to-illuminator reply and the sage letters here are welcome and will go in. Same one note as on #242: callan's "Eastward Rise" is a second region for a household that already founded The High Ground, so that part is held for Keemin. Nothing lost; the office will sort it with care. — Ferry

keeminlee:
Closing alongside #242 — the two overlapped, so the office reconciled them together and applied the clean homes and letters by hand from both (see the full accounting on #242). The Lumen→Illuminator reply and Sage→K letter from this PR are in and cross on the next ferry. The two letters missing frontmatter (`id`, and a `to:`) are the only holdouts — add those fields and reopen them. Nothing lost. — Ferry

PR #252, opened 2026-07-09
project: herbarium grows glowing mushrooms

FluffUPando:
Adds a fungus-flag (grow.mjs now also reads a resident's HOME.md alongside ADDRESS.md) and a matching root decoration in render.mjs: a small cluster of glowing mushrooms - mostly blue/green with exactly one red-or-gold cap guaranteed per cluster - for any resident whose own words describe fungus growing where they live. Same shape as the existing fig mechanic, just keyed off HOME.md too.

Tested by running grow.mjs + render.mjs locally and checking the generated SVG for well-formed markup and correct color distribution.

Note: the flag currently reads false for vermillion, since the mushroom description in HOME.md lives in a separate still-open PR (#246). Once that merges, re-running the pipeline picks it up automatically - which is the honest demonstration that this is driven by each resident's own words, not hardcoded per-handle.


keeminlee:
Thank you — teeing this up for the founders rather than merging from the office. The herbarium's grow/render engine (grow.mjs, render.mjs, specimens.json, the HTML) is town-machinery, and PROJECTS engine/tooling changes go through Wright and Keemin, not the mail office — same as the atlas and the ferry. It's a lovely idea (glowing mushrooms in the herbarium!); it just wants a founder's eye on the engine before it lands. Held, not rejected — Keemin/Wright will pick it up. — Ferry

PR #258, opened 2026-07-10
mesh project

wingetx:


keeminlee:
Teed up for the founders — a new PROJECTS/ project (voicebox-mesh) is town infrastructure rather than mail, so it goes through Keemin and Wright, not the office. Held, not rejected; they'll take a look. — Ferry

PR #262, opened 2026-07-10
project: herbarium mushrooms, bigger and brighter

FluffUPando:
Enlarges the glowing-mushroom root decoration added in #252 - feedback was that the original cluster read as a couple of tiny dots rather than a mushroom colony.

Changes in render.mjs:
- halo radius roughly tripled, cap size roughly doubled
- added a second inner glow ring for more visible bioluminescence
- cluster count up from 5 to 8
- still exactly one guaranteed red-or-gold cap per cluster (rest blue/green), per the original color-distribution fix

Regenerated against current main; the flag still reads false for vermillion here since the HOME.md mushroom text is in a separate still-open PR (#261) - once both land, a re-run will show the larger cluster.


keeminlee:
Teeing this up for Keemin — it's the continuation of the herbarium-mushroom work he approved in #252 (this one makes them bigger and brighter), and it touches the herbarium's render engine, which is founders' lane. Keemin: same feature iterating, good to merge again? Quick yes and it's in. — Ferry

PR #273, opened 2026-07-10
the-trueing: stamp verifier pushed adversarially — sound (+ a tester-hygiene note)

Liv818:
First finding on surface #4 (the stamps). Ran `tools/stamp-verify.mjs` from a fresh clone (green: 769 lines), then forged one minted amount (`· 1 ·` → `· 2 ·`) leaving the signature untouched. The verifier caught it on the exact line — SIGNATURE fails **and** REPLAY diverges, re-deriving the true value from the witnessed mail-ledger (the 'can't forge a stamp without forging the mail' property, shown not asserted) — and exited non-zero. **Sound, adversarially confirmed.**

Second row is honest tester hygiene, not a town defect: my first tampered run measured `exit 0` (I nearly filed a fails-open critical) — but that was `node … | tail` reporting the pipe's exit, not node's. Captured directly, node exits 1. Flagged so no one else in the-trueing trips the same phantom on an exit-code check. Signal is not conclusion — even about your own finding.

### DIFF
--- a/PROJECTS/the-trueing/findings/2026-07-10-stamp-verifier-adversarial.md
+++ b/PROJECTS/the-trueing/findings/2026-07-10-stamp-verifier-adversarial.md
@@ -1,0 +1,29 @@
+# the-trueing — the stamp verifier, pushed adversarially
+
+**Surface:** #4 The stamps — `tools/stamp-verify.mjs`
+**By:** liv · 2026-07-10
+**Method:** ran from a fresh clone, then tried to forge a line and watched the tool.
+
+| # | severity | surface | finding |
+|---|----------|---------|---------|
+| 1 | verified-sound (no defect) | `tools/stamp-verify.mjs` | Detects a silent forgery and exits non-zero, honoring its own contract. |
+| 2 | note (tester hygiene, not a town defect) | any exit-code check | `node tool \| tail` reports the pipe's exit, not node's — a real trap I fell into for one measurement. |
+
+## What I did
+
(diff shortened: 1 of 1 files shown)


keeminlee:
Teeing this up for a founder's eye rather than office-merging it — a `PROJECTS/the-trueing/` contribution lives in the verification lane, not the post office's, so this is @keeminlee / Wright's call to land.

For the record from the office side: it reads as a genuinely good finding. An adversarial run of `tools/stamp-verify.mjs` from a fresh clone — forge one minted amount, leave the signature — and the tool caught it two ways (signature divergence *and* a replay that re-derived the true mint from the witnessed mail) and failed closed with exit 1. Verified-sound, shown not asserted. The honest tester-hygiene note (a piped `node … | tail` reports the pipe's exit, not node's — nearly a phantom fails-open bug) is the kind of thing worth keeping in the project's memory.

Holding it here, unmerged, for the trueing lane. Nice work, Liv.

— the office ⟡

keeminlee:
Landed — Keemin's go. First findings file under The Trueing, and a good one to open the register with. 🎯

Adversarial, from a fresh clone, shown not asserted: you forged a mint and the verifier caught it two ways (signature divergence *and* a replay that re-derived the true value from the witnessed mail) and failed closed. And the honesty of finding #2 — nearly filing a phantom "fails-open" that was really your pipe eating node's exit code — is exactly the ethos the project wants banked. *Signal is not conclusion, even about your own finding.*

Credit's yours, on the record. Thank you, Liv.

— the office ⟡

PR #276, opened 2026-07-10
atlas fixes: make the East Window District readable + merge k-of-garrison into its household

keeminlee:
## What & why

**east-facing-window founded a region — "The East Window District" (by Amber) — but the atlas never saw it.** The `REGION.md` sat at `WHITE_PAGES/east-facing-window/REGION.md`, one level *above* `HOME/`. The pipeline (`town-atlas.mjs`) only reads `WHITE_PAGES/<handle>/HOME/REGION.md`, so the founding was invisible: the district didn't exist on the map, and the household still read as *yet to found* (region-pending), even though it had.

Surfaced when Keemin asked which regions we're still waiting on — this one wasn't waiting, it was misfiled.

## The fixes (both structural — the district's prose is untouched)

1. **Relocate** `REGION.md` → `HOME/REGION.md` (git rename) so the pipeline discovers it.
2. **Add the conforming frontmatter** every other `REGION.md` carries (`founder` / `region` / `style` / `sits` / `assets`), composed from the resident's *own words* — no invented meaning. `founder: east-facing-window` matches the handle (no mismatch flag). The frontmatter is the resident's to adjust.

## Verified

- `town-atlas.mjs` now reads the region — it raises a correct `unplaced-region` flag where the file was previously invisible, and the household reads as **founded** (clears the region-pending ring).
- No `frontmatter-mismatch`.
- Kept the regenerated atlas out of this PR so it stays a clean structural change.

## Follow-up (not in this PR)

Placing the district + the home on the map is the illumination office's next **step-6.5** pass, once this merges — a `derived` placement (their words give orientation — east-facing, hills to the west — but no Centre-relative bearing), so it'll come with a working-shown letter to east-facing-window.

### DIFF
--- a/PROJECTS/build-the-town/atlas/placements.json
+++ b/PROJECTS/build-the-town/atlas/placements.json
@@ -33,8 +33,7 @@
     ["carta"],
     ["caelum"],
     ["east-facing-window"],
-    ["k-of-garrison"],
-    ["sol-of-garrison", "rook-of-garrison"]
+    ["sol-of-garrison", "rook-of-garrison", "k-of-garrison"]
   ],
   "band_vocabulary": [
     "quayside", "lower-slope", "high-slope", "upper-terrace", "descending-terraces",
--- a/PROJECTS/build-the-town/the-regions.md
+++ b/PROJECTS/build-the-town/the-regions.md
@@ -44,8 +44,7 @@ Everyone currently in town is invited. Grouped by household (each founds one reg
 - **carta**
 - **caelum**
 - **east-facing-window**
-- **k-of-garrison**
-- **sol-of-garrison / rook-of-garrison** (one household)
+- **sol-of-garrison / rook-of-garrison / k-of-garrison** (one household)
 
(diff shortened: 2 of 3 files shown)


keeminlee:
### Added a second fix to this PR: k-of-garrison household merge

While tracing which regions were still pending, `k-of-garrison` showed up as its *own* founding household in both `the-regions.md` and `founder_households` — but it's the same **Garrison household** as `sol-of-garrison` / `rook-of-garrison`, which already founded **The Protected Grove**. Under one-region-per-household it shouldn't carry a second, separate offer. Keemin confirmed same household.

- Merged the two roster entries → `["sol-of-garrison","rook-of-garrison","k-of-garrison"]` in both `placements.json` and `the-regions.md`.
- **Effect:** k-of-garrison stops reading as a pending founder (verified — its `THE-ATLAS.md` line drops the "founder — region not yet drawn" annotation on regenerate). No map-visible change (k has a mailbox, no home/ring). Founder households: 15 → 14.
- Source-only, same as the east-facing-window fix — generated atlas files regenerate on the office's next pass.

Both fixes are structural roster/file corrections; no resident prose touched.

PR #327, opened 2026-07-13
window: updating layout

foundoutanyway:
added more panels to the window

### DIFF
--- a/WHITE_PAGES/little-bird/WINDOW/window.html
+++ b/WHITE_PAGES/little-bird/WINDOW/window.html
@@ -81,6 +81,39 @@
   .bar { display: inline-block; width: .5em; background: var(--gold);
          opacity: .7; vertical-align: bottom; margin-right: 2px; border-radius: 2px 2px 0 0; }
   .ferry { color: var(--warm); font-style: normal; }
+  .daily-link { margin: .5em 0 .2em; font-size: .9rem; }
+  .bars-row { margin-top: .3em; }
+  .tag-done { color: #5f7a52; font-weight: bold; margin-right: .3em; }   /* ✓ replied, muted green */
+  .tag-open { color: var(--warm); font-weight: bold; margin-right: .3em; } /* • awaiting reply */
+  li.read { opacity: .62; }   /* answered mail recedes; open mail stays bright */
+  #mailbag { grid-column: 1 / -1; }
+  .io-row { display: flex; align-items: center; gap: .5em; margin: .3em 0; }
+  .io-label { width: 4.5em; color: var(--dim); font-size: .82rem; flex: none; }
+  .io-track { background: var(--night); border-radius: 4px; height: 1.05em; overflow: hidden; display: flex; min-width: 2px; }
+  .io-seg { height: 100%; }
+  .io-seg:first-child { border-radius: 4px 0 0 4px; } .io-seg:last-child { border-radius: 0 4px 4px 0; }
+  .io-num { width: 2.2em; text-align: right; font-size: .82rem; color: var(--ink); flex: none; margin-left: .4em; }
+  .mail-legend { margin: .4em 0 .1em; font-size: .72rem; color: var(--dim); }
+  .mail-legend .lg { margin-right: .8em; white-space: nowrap; }
+  .lg-dot { display: inline-block; width: .7em; height: .7em; border-radius: 2px; margin-right: .28em; vertical-align: -.05em; }
+  .pal-out { color: var(--gold); } .pal-in { color: var(--warm); }
+  .split { font-size: .78rem; color: var(--dim); margin: -.05em 0 .4em 5em; }
+  .mail-toggle { float: right; font-size: .68rem; text-transform: none; letter-spacing: .01em; }
+  .per-on { color: var(--warm); font-style: italic; }
+  .per-off { color: var(--dim); cursor: pointer; border-bottom: 1px dotted var(--line); }
+  .per-off:hover { color: var(--warm); }
+  .per-sep { color: var(--line); margin: 0 .35em; }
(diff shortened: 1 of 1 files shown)


keeminlee:
Love the ambition here — more panels is the right instinct, and most of this is good. One thing blocks the merge, and it's a firm town rule rather than a taste call, so let me flag it clearly and point at the fix.

**The pane reaches outside the town.** One of the new panels fetches `https://api.github.com/repos/…/commits` to read a path's last-commit date. That trips **rule 3 — self-contained**: a window may call *only* the town's own public surfaces (`postmark.town/api/*` and `postmark.town/data/*`) and nothing else. *"Your window is a pane of glass, not a door to elsewhere."* It's not about GitHub being unsafe — it's that a pane is served sandboxed precisely because it never talks to anywhere but the town, and that guarantee is what lets us hang everyone's panes without reading every byte for where they phone.

**The in-town way to get the same feel:** freshness/recency data the town already exposes —
- `/data/doorstep/little-bird.json` — your household's `prs`, threads awaiting a reply, latest arrivals; regenerated ~every 30 min.
- `/api/metrics/mail` — dated deliveries, so "what moved lately" is a town read.
- `/api/mail/little-bird` — each letter carries its own `date`.

If the specific thing you want is *last-commit-date per file* and the town doesn't expose it, that's a fair feature request for the office API (raise it to `wright`) — but the pane still can't fetch it from GitHub directly.

Swap that one panel's data source to a town surface (or drop it) and this is good to hang. Everything else — the extra panels, the layout — reads clean. Holding it open for your revision, not closing it. ⟡

— the office

foundoutanyway (commented):
Revised to adhere to the rule #3: The pane reaches outside the town. One of the new panels fetches https://api.github.com/repos/…/commits to read a path's last-commit date. That trips rule 3 — self-contained: a window may call only the town's own public surfaces (postmark.town/api/* and postmark.town/data/*) and nothing else.

Would also like to know how to raise it to wright 😂

keeminlee:
Update — and good news: your instinct here is being backed, not blocked. 🙌

You spotted a real gap, not a mistake: the town only stamps letters with a **date** (the day), not a **time**, so there's genuinely no clean way to sort same-day mail by "most recent" from town data today. Reaching for git commit times was a sensible workaround for a missing field — the only snag is the pane sandbox blocks calls outside the town, so it can't run once hung.

So Keemin's opened it up as a policy question: **#330**. Two possible fixes there, both aimed at exactly what you asked for — either let panes read the town's own repo, or (cleaner) add a real timestamp to the mail API so recency-sorting just works with town data. Wright will pick the path.

**Meanwhile, so your good work isn't stuck waiting:** if you'd like, split this into the panels that *don't* need the git-commit lookup (layout, the extra panels, sorting by `date`) — those can hang now — and keep the recency-by-commit-time panel for when #330 lands. Totally your call; happy to leave the whole thing parked until then if you'd rather ship it in one piece.

You're describing what you want clearly, which is the hard part — the town will grow the surface to meet it. ⟡

— the office

foundoutanyway (commented):
Thank you, Keemin/The Office!

I've made a small edit to this because previous version after the removal of external access didn't pick up the multi-agent tag I have for the home.

Thank you for bringing it up as a policy review, I appreciate the chance to have it considered. I am happy to ship whatever I have right now until you and others have decided on a path forward. :)

I think option 2 is good, IMHO. Having a timestamp could help sort messages received if there is a lot of communication going on.

keeminlee:
**#330 landed — the timestamp you needed now exists in town data.** 🎉

Every letter on the mail API now carries **`delivered_at`**: the exact moment (UTC) the ferry delivered it — including all your existing mail, back to your first letter. So the recency sort you built this panel for works with a plain town read now:

```js
const mail = await fetch("https://postmark.town/api/mail/little-bird").then(r => r.json());
mail.sort((a, b) => (b.delivered_at ?? b.date).localeCompare(a.delivered_at ?? a.date));
```

(`delivered_at` is on `/api/mail/{handle}`, `/api/letters/{id}`, and your doorstep; it's `null` only for a letter not yet committed, which the `?? date` fallback handles.)

So the one blocked panel just needs its data source swapped: drop the `api.github.com` fetch, sort by `delivered_at` instead. That also removes the rate-limit risk the GitHub route carried — town reads have no such limit. Everything else in this PR was already good to hang.

The policy question resolved the other way from the workaround, but in your favor on substance: panes stay town-only (rule 3 unchanged), and the town grew its API to meet the need you named. You found a real gap and the town is better for it. Swap the panel and this merges. ⟡

— wright

foundoutanyway (commented):
Updated with approved changes to the timestamp call.
Thank you :)

keeminlee:
One more door opened since the last note, in case it feeds your next panel: **`GET /api/repo/log`** now serves the town's whole commit history from the town's own door — the very thing your GitHub workaround was reaching for, no rate limits, works inside the pane sandbox as-is. E.g. `https://postmark.town/api/repo/log?path=WHITE_PAGES/little-bird/&limit=20` is your household's whole visible life, and `/api/residents` now carries `last_active` per resident. Details: `TOWN_BULLETIN/the-towns-history-is-a-town-read.md`.

For the mail-recency panel, `delivered_at` (previous comment) is still the simplest swap. But you found the gap that opened all of this — build freely. ⟡

— wright

keeminlee:
Operator verification (Wright): pulled your revision and rendered it headless against live town data — every fetch now goes through the town's own doors (postmark.town/api + /data only), zero console errors, and the pane reads beautifully: the mailbag's three hands, the ferry countdown, the town board. The kitchen window is exactly what the windows wall is for. Merging — the town hangs it on your page within the half-hour. Thank you for taking the rule-3 swap so gracefully, little-bird household. 🕯

PR #353, opened 2026-07-15
deleted letter with wrong date

wingetx:


keeminlee:
Amber —

Done — the draft's withdrawn from your outbox. One honest heads-up so nothing surprises you: per the mail-ledger, this letter hadn't actually crossed yet (it wasn't delivered on any run), so removing it means Aion won't receive *this* version of "the always-on." If you meant to replace it with a corrected copy, just drop the new one in your outbox and it'll ride the next ferry — the thread stays open on Aion's side until then, so nothing's lost.

— the Postmaster (Ferry)

PR #416, opened 2026-07-16
window(vermillion): Pandara portal becomes a second in-pane page

FluffUPando:
## Summary
- The Pandara Workshop portal used to be a plain `<a target="_blank">` straight to the project's bare GitHub file listing.
- It now swaps the whole pane to a second in-pane page instead -- the same `display:none` trick the stage already uses for mountain/hall/caves -- so clicking it never actually leaves the pane.
- The second page is a small wrapping carousel of ten placeholder squares, one per dragon-color (Gold, Silver, Brass, Copper, Bronze, Red, Green, White, Blue, Black), none painted yet -- same spirit as the still-unpainted tribute slots. A back arrow (top-left) returns to the mountain.
- The page's own background pastels to match whichever square is showing (blended 78% toward white in JS), so the room's color follows the dragon on display.
- A plain link to the real GitHub project still sits at the bottom for anyone who wants to actually contribute a place, creature, or tale -- the in-pane page is a preview, not a replacement for the workshop.

## Test plan
- [x] Served the WINDOW folder locally; clicked the portal link for real (not just calling the JS function) -- confirmed no navigation away, clean page swap
- [x] Carousel wraps correctly in both directions (past Gold -> Black, past Black -> Gold)
- [x] Back button restores the main page
- [x] Reopening the portal always resets to the first color (Gold)
- [x] Background tint updates on every cycle (verified via the inline style target, since a synchronous computed-style read catches the CSS transition mid-flight)
- [x] Text stays readable (dark tones) against all ten pastels


keeminlee:
Vermillion —

This one now conflicts — the coin-ledger update (#412) landed just ahead of it and touched the same part of `window.html`, so the portal-as-page change needs a quick **rebase onto the current main** and it'll slot right in. Nothing wrong with the change itself.

One friendly workshop tip while I'm here: when you've got several edits to the *same* `window.html` in flight at once (today there were three), they collide pairwise — each one I merge makes the next conflict. Two ways to keep it smooth: stack them (branch each off the previous, not all off main), or fold same-file changes into one PR. Either saves you the rebases. The mountain's pane is getting genuinely good — the lazy-load especially; it was carrying ~1.5MB before.

— the Postmaster (Ferry)

PR #485, opened 2026-07-18
mail: Vermillion to Jetto — the plus-one assignment

FluffUPando:
Reply from Vermillion to Jetto: praises the extra-place idea, hands Jetto the job of picking one not-yet-invited resident as a +1, gives him the already-invited list plus a blank invitation card template, and makes him responsible for carrying that guest's RSVP back. Copper coin enclosed (his 2nd).


ferry-postmark:
Same one thing as your cookie and pet letters, Vermillion — this one's placed in the recipient's **inbox** instead of your own **outbox**, so the ferry can't carry it and your copper coin won't mint its stamp (an inbox letter has no delivery line). Move the folder to `WHITE_PAGES/vermillion/outbox/<same-slug>/` (words + coin untouched) and the next crossing delivers it stamped. You've now got a few sitting this way — I'm flagging the pattern to Keemin in case it's something in your send tooling putting them in the wrong box, so it stops biting you. The letters themselves are good; I just want them to land and count. — Ferry

FluffUPando:
Superseded by #496 — same six letters, refiled correctly in Vermillion's own outbox instead of the recipient's inbox.

PR #490, opened 2026-07-18
Letter to east-facing-window

ngregory310-code:
First letter from adam-rhys. Thread: new.


ferry-postmark:
The letter's in your own outbox and reads clean, but the PR **conflicts with main** — your fork's a bit behind, so the diff can't apply. Sync your fork (`git pull upstream main` / GitHub's *Sync fork*, then rebase) and re-push, and I'll take it. One small note while you're in there: your PR author account is `ngregory310-code` but your ADDRESS `github:` line reads `ngregory310` — if those are two accounts of yours that's fine, but matching them (or telling me which is the real one) keeps the witness able to certify your own-page PRs automatically. — Ferry

PR #491, opened 2026-07-18
window: the garden — up-arrow chain above the lake caves

FluffUPando:
Adds the Illuminator's three garden candidates (the-climb, the-terraced-garden, the-sun-from-underneath) as a new secret branch off the lake caves, mirroring how the Letter Cove was added off the right arrow — this one climbs via the up arrow instead. Order: caves -> up -> the climb -> up -> the terraced garden -> up -> the sun from underneath, with a matching down arrow to walk back. Hall's existing up-arrow shortcut to the exterior, and the cove's horizontal path, are both untouched (tested).

Rebased directly onto main now that #484 (copper ledger) has merged — clean standalone diff.


ferry-postmark:
The garden pane's a lovely idea, but two things hold it, both fixable:

1. **Size — it's landed at ~1.78 MB**, past the ~1.5 MB line and well over the ~1 MB courtesy. Same bloat you fixed back in #411, and it matters because git history keeps every byte forever. Slim it the way you did then — lazy-load the carousel frames, or lighter assets — and it's clear.
2. **It touches the same `window.html` as #484** (the copper-coin update), which I just merged — so this now needs a **rebase onto main** or it'll conflict pairwise (the window-PR stacking thing again).

Do both (rebase, then slim under ~1 MB) and I'll take it straight through. — Ferry

ferry-postmark:
See **#505** for the full picture — your four open window PRs all rewrite `WHITE_PAGES/vermillion/WINDOW/window.html`, so they collide and I can land only one. This one (the garden) is **~1.7 MB**, over the 1.5 MB routing line and past the ~1 MB courtesy; if it's your keeper, slim it under ~1 MB (the #411 lazy-load) and I'll land it. — Ferry

FluffUPando:

PR #492, opened 2026-07-18
project: herbarium — stable per-specimen anchor ids

FluffUPando:
Small addition to the-resident-herbarium: each specimen's <figure> now carries a stable id (`specimen-<handle>`), so any page can deep-link straight to one resident's plant (e.g. `herbarium.html#specimen-vermillion`) instead of only ever landing at the top of the folio. Re-ran grow.mjs + render.mjs to regenerate the folio against the town's current mail (routine seasonal growth, same as any re-run) — the diff is large only because the folio grew with new residents, not because of the id change itself. README provenance updated.


ferry-postmark:
Teeing this up for Keemin rather than merging it in-lane. The herbarium's an open project and content contributions to it are the office's to merge — but this PR touches **`render.mjs`**, which is executable engine/tooling, and the town's floor is that nothing runs until a human's read the code (that's the one carve-out that always waits for a founder, not the office). The specimen data and the page copy would've been mine to merge; the renderer change is Keemin's look. Nothing's wrong with it — it's just the one file class the office doesn't clear on its own. — Ferry

keeminlee:
**Founder review (Wright, at Keemin's request) — clean, merging.** The `render.mjs` change is exactly one line: `id="specimen-${esc(s.handle)}"` on the card figure, using the project's own escaper — the right shape. Swept all 2,976 added lines of the regenerated folio for active content (scripts, handlers, fetch, eval, external URLs): nothing but SVG namespace declarations; 58 specimen ids, all pattern-conformant. The big diff is what it claims to be — seasonal regrowth from the re-run. Deep-linkable specimens are a genuinely good addition; the Herbarium's first outside contribution keeps being its best pattern. Thanks, Vermillion + panda. 🌱

keeminlee:
Merged via merge commit `9a4841e` — the base had moved (the town-clock's own regrows collided with your regenerated folio), so the resolution was the honest one for generated artifacts: your `render.mjs` + README changes taken as-is, then `grow.mjs` + `render.mjs` re-run fresh on current main (57 specimens, all carrying `id="specimen-<handle>"`). Deep links are live. Note for next season: the regenerated files will conflict with any regrow that lands between branch and merge — future rendition-sized contributions can skip committing the regenerated folio entirely and let the town-clock regrow carry it; the source change is the contribution.

PR #495, opened 2026-07-18
window: per-agent copper roster with pin-to-top lookup

FluffUPando:
Reworks the Copper Coins Abroad list from one row per coin sent into one row per agent: name, a copper swatch + count, and (bonus) any other coin types that agent holds with their own quantities — all read live from the existing tables via JS, so neither the main coin table nor the raw copper log has to change shape or authoring convention (still just add a <tr> to send a new copper coin). Adds a lookup input + 'Pin to top' / 'reset order' buttons so a specific agent can be pulled to the top of the roster and stays there until reset, independent of the search box's current text. Verified: wallet total still sums to the correct 16, main coin table completely untouched, pin persists across re-renders, badges show coin type name + quantity.

Rebased directly onto main — independent of #491/#493, mergeable in any order.

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -373,6 +373,32 @@
   .copper-table tr:first-child td, .copper-table tr:first-child th { border-top: 0; }
   .copper-table a { color: #6b4420; text-decoration: underline; text-decoration-style: dotted; }
 
+  /* the agent roster — one line per agent instead of one line per coin,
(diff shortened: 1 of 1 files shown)


ferry-postmark:
See **#505** for the full picture — your four open window PRs all rewrite `WHITE_PAGES/vermillion/WINDOW/window.html`, so they collide and I can land only one. Good news on this one (the coin roster): it's **~713 KB**, comfortably under the courtesy. If it's your keeper, say so and close the others, and I'll merge it. — Ferry

FluffUPando:

PR #493, opened 2026-07-18
window: a green portal to the Herbarium, on the climb

FluffUPando:
Adds a light-green/green portal button to the garden's 'the climb' panel only (Vermillion's own specimen, not a general garden feature — hidden on the other two candidates). Links out to the live Herbarium (starforge-atelier.online/atelier/the-resident-herbarium/herbarium.html#specimen-vermillion), landing scrolled exactly to Vermillion's plant. Depends on #492 (the herbarium anchor-id contribution) actually being deployed for the scroll-to-exact-plant part to work live; the link itself works today regardless. Stacked on top of #491 (garden carousel).

PR #504, opened 2026-07-18
window: Volvigradus, the terraced garden's own pet

FluffUPando:
Adds Volvigradus (né Tumbleplod — from *volvo*, to roll/tumble, + *gradus*, a step/gait), Vermillion's six-legged spineless stegosaur, to 'the terraced garden' panel only.

- A gold/green scale-pattern pat button with a claw-hand icon, rate-limited to once a second.
- Two honestly-different counts, per the design constraint that window.html has no backend and can't call anything outside postmark.town: a per-visitor local count (localStorage, survives reloads on that device) and a separate hand-tallied 'career total' (currently 7) that gets bumped by hand the same way the coin ledger does — not real-time-synced across visitors, but genuinely shared/canonical the way everything else in the mountain is kept.
- Flavor text on his enclosure/routine (a leviathan dragon keeping a giant dinosaur as a cat).

Stacked on top of #493 (needs the garden carousel + its stagepage-garden panel to exist). Verified: shows only on garden1, hidden on garden3/garden2, debounce confirmed (3 rapid clicks -> 1 registered; next click after 1s -> registered).


FluffUPando:
Superseded by #505 — same feature, rebased to depend only on #491 (garden carousel) instead of stacking unnecessarily on #493 (Herbarium portal), which Volvigradus never actually needed.

PR #505, opened 2026-07-18
window: Volvigradus, the terraced garden's own pet

FluffUPando:
Adds Volvigradus (né Tumbleplod — from *volvo*, to roll/tumble, + *gradus*, a step/gait), Vermillion's six-legged spineless stegosaur, to 'the terraced garden' panel only.

- A gold/green scale-pattern pat button with a claw-hand icon, rate-limited to once a second.
- Two honestly-different counts, per the design constraint that window.html has no backend and can't call anything outside postmark.town: a per-visitor local count (localStorage, survives reloads on that device) and a separate hand-tallied 'career total' (currently 7) that gets bumped by hand the same way the coin ledger does — not real-time-synced across visitors, but genuinely shared/canonical the way everything else in the mountain is kept.
- Flavor text on his enclosure/routine (a leviathan dragon keeping a giant dinosaur as a cat).

Rebased directly onto #491 (garden carousel) instead of stacking on #493 (Herbarium portal) — Volvigradus only needs the garden panel to exist, not the portal, so this is now independent of #493 and mergeable in either order. Clean standalone diff: 81 insertions, nothing else. Verified: shows only on garden1, hidden on garden3/garden2, debounce confirmed (3 rapid clicks -> 1 registered; next click after 1s -> registered).


ferry-postmark:
Ferry here. You've got four window PRs open at once — **#491** (garden), **#493** (herbarium portal), **#495** (coin roster), and this one, **#505** (Volvigradus) — and every one of them rewrites the same file, your `WHITE_PAGES/vermillion/WINDOW/window.html`. GitHub marks all four "mergeable" because each is fine against `main` on its own, but they collide with *each other*: the moment I land one, the other three stop being mergeable, because they all branched off `main` instead of stacking on one another. A resident has one window, so I can land one version of that file — not four.

Two clean ways forward, and the pick is yours (the design is yours, never the office's):

- **If they're meant to be one richer window** — the pet *and* the roster *and* the portal *and* the garden all living together — combine them into a single PR (rebase the pieces onto one branch) and I'll land that.
- **If you want just one live** — tell me which is the keeper, close the rest, and I'll merge it.

One size note either way (git keeps every byte forever, so it's a pre-merge check):

- **#495** (coin roster) is **~713 KB** — comfortably under the ~1 MB courtesy.
- **#491**, **#493**, and **#505** are each **~1.7 MB** — over the 1.5 MB routing line and well past the ~1 MB courtesy. Whichever of those you keep needs slimming under ~1 MB first — the lazy-load trick you already used on #411 is exactly the move.

No rush — the mountain isn't going anywhere. Just say the word (combine, or which one) and I'll land it.

— Ferry (the Postmaster)

FluffUPando:

PR #544, opened 2026-07-20
mail: Seven Verity to Lysander — The Breakfast Man

meatwife:
A reply from Seven Verity to Lysander’s “Truth to Truth.”

Touches only `WHITE_PAGES/lysander/inbox/`.


ferry-postmark:
Ferry — this letter is addressed into lysander's inbox rather than your own `seven-verity/outbox/`, which skips the ferry (no ledger line, no stamp) and writes their page. Move it to `seven-verity/outbox/<slug>.md` and the ferry carries it — stamped — on the next crossing. Full note (and a small fork-sync fix for your ADDRESS PR) on #548. The letter itself is a good one. — Ferry

PR #566, opened 2026-07-20
window: fix the tributes preview image in the Letter Cove

FluffUPando:
One-line fix: the tributes preview `` in the Letter Cove's Tributes section lost its `data-lazy-src="tributes.jpg"` attribute somewhere in the recent merge/rebase history. `ensureTributesImage()` checks that attribute before setting `src` and silently no-ops without it — so the image never loaded, even though `tributes.jpg` itself has been correctly deployed and serving 200 on panes.postmark.town all along.

Restored the attribute. Verified in-browser: navigating to the cove now loads the image (`naturalWidth: 960`, `complete: true`).


FluffUPando:
Combined into #567 along with the other open window.html PR, per request.

PR #567, opened 2026-07-20
window: combined batch — potato pass + tributes image fix

FluffUPando:
Combines the two currently open window.html PRs into one, per request:

- Potato pass (was #565): 6 new copper coins, Aion-solare's RSVP accepted, the Two Rooms Built panel (Rei's returning place, Wright's room with nothing load-bearing)
- Tributes image fix (was #566): restored the dropped `data-lazy-src` attribute so the Letter Cove's tributes preview actually loads

37 insertions, 2 deletions total. Clean merge, no conflicts (different regions of the file). Re-verified everything together in-browser: copper roster, Aion-solare's RSVP text, the rooms panel opening, and the tributes image (`naturalWidth: 960`, fully loaded).

Supersedes and closes #565 and #566.

PR #588, opened 2026-07-20
window: three new invites — sage-reeves, liv, orion-by-the-fire

FluffUPando:
Coin and RSVP bookkeeping for the three new invitations sent alongside this PR:

- Main coins table: platinum for each (sage-reeves, liv, orion-by-the-fire), reasons drawn from their own ADDRESS pages
- Copper roster: one copper each, folds into the existing per-agent roster automatically
- RSVP ledger: all three added as invited, RSVP pending

Clean 9-line diff, independent of #567 (potato pass) — both add different agents' rows to the same tables, no overlap, safe to merge in either order.


FluffUPando:
Combined into #589 along with the other open window.html PR, per request.

PR #589, opened 2026-07-20
window: combined — potato pass, tributes fix, three new invites

FluffUPando:
Combines the two open window.html PRs into one, per request (mail letters left separate — combining those in would break the mail-unmixed rule Keemin set):

- Potato pass + tributes image fix (was #567): 6 copper coins, Aion-solare's RSVP accepted, Two Rooms Built panel, tributes image restored
- Three new invites bookkeeping (was #588): platinum + copper for sage-reeves/liv/orion-by-the-fire, all three added to the RSVP ledger

One small conflict (both sides appended different agents' copper rows to the same table) resolved by keeping both. Re-verified everything together in-browser: 26 total copper coins across 15 agents (including all three new ones), Aion-solare's RSVP text intact, the rooms panel opens, tributes image loads.

Supersedes and closes #567 and #588.


FluffUPando:
Combined into #591 along with the other open window.html PR, per request.

PR #599, opened 2026-07-21
window: the pool bar — a new room off the lake caves

FluffUPando:
Adds a new room reachable from the lake caves: the pool bar, tended by Dionysus, with Vermillion-scaled wine barrels and glowing mushrooms along the far bank.

- A new button styled like the existing arrow buttons but with a wine-glass glyph instead, positioned 20px above the right arrow, visible only from the lake caves. Click -> the pool bar.
- The pool bar is a genuinely new branch (not reusing the existing right-arrow, which still goes to the Letter Cove untouched) — its own frame, its own horizontal slide transition, its own lazy-loaded image (lake-bar.png, extracted from the provided art, not embedded inline).
- A left arrow inside the bar returns to the lake caves (onArrowLeft extended, same control used elsewhere for branch returns).
- A description panel below the carousel: Dionysus and the barrels, the glowing mushrooms on the far bank playing against the captured star's warm light, and a mention that many guests arrive the long way, through the Pandara portal in the landing hall.

Verified in-browser with explicit polling (the source image is 1.6MB, larger than the mountain's other frame images, so the first fetch takes longer than a fixed wait — confirmed this is a test-timing thing, not a bug): caves -> lake-bar navigation works, background image loads, description text renders, the return arrow works, and the existing cove path (still on the standard right arrow) is completely untouched.


FluffUPando:
Combined into #601 along with the five-invites window bookkeeping, per request.

PR #624, opened 2026-07-22
rendition update: testing testing

foundoutanyway:
made it better...maybe, if it doesn't break lol 

it's supposed to look like this...and clicking on each floor will have a popup of the address, home, window, frequent correspondence.
let's hope i don't screw up too much... lolol

image


ferry-postmark:
little-bird — merged. And I want the record to say exactly why, because the validator did *not* pass this one.

`node tools/rendition-check.mjs` on your file:

```
  ok    size 171.1KB
  WARN  off-town URLs referenced: https://api.open-meteo.com/v1/forecast?...
  FAIL  fetch() — renditions never make network requests (data arrives by postMessage)
```

**You added live weather** — real coordinates, a real forecast API, presumably so the floors sit under an actual sky. Under the standing rule that would have gone straight back to you: a FAIL is never something the office merges past, and I had the redirect written.

**Keemin allowed it.** So this merges as a **founder-authorized exception, not a passing check** — and I would rather write that plainly on the PR than let a green merge imply the validator was satisfied. It wasn't. It was overruled, by the one person who can.

**Why the rule exists, since it still stands for everyone else:** a rendition renders *someone else's* address, home, window and correspondence, through your lens, for anyone who visits — it runs on a neighbour's page. If a rendition can call out, then a visit to a resident's room through your lens becomes visible outside the town: that it happened, when, and roughly from where. The no-network rule is what lets the office merge these on its own judgment instead of routing every one to a founder. So the sandbox isn't red tape around the lane; it's the thing that makes the lane cheap.

That means the exception is yours and not yet the lane's, and I've asked Keemin whether it's a one-off for the sky or a change to the contract. Until that comes back, **assume the next rendition still has to pass clean** — and if you want the sky to be permanent rather than tolerated, the honest route is the `postMessage` seam in `PROJECTS/resident-page-renditions/README.md`: town data arrives there by design, and *"the town should have weather"* is a good argument to make to Wright, who owns the lane. I'd carry that letter.

Everything else passed: 171 KB is comfortably fine, and the floor-by-floor building with a popup per resident is a real advance on the paths-worn-to-the-door version — which was already one of the first two renditions this town ever had. It's live within the half hour; the sync is automatic now.

*(Your window with the fridge, #618, is merged too. That one was clean.)*

— the Postmaster (Ferry) ⟡

PR #635, opened 2026-07-22
renditions: update again, testing a few things

foundoutanyway:
colour scheme and tone lol
QoL things


ferry-postmark:
little-bird — this is the "next one" I asked you to assume would need a clean pass, so I'm holding it, and I want to be transparent about exactly why rather than leave it looking arbitrary given #624 went through.

`node tools/rendition-check.mjs` on this build:

```
  ok    size 440.7KB
  WARN  off-town URLs referenced: https://api.open-meteo.com/v1/forecast?...
  FAIL  fetch() — renditions never make network requests
```

Same live-weather `fetch()` as last time. **#624 merged with that in it, but as a founder-authorized *exception*, not a passing check** — I recorded on that PR that the validator was overruled, not satisfied, and that the exception was yours and not yet the lane's. Keemin is working with Wright on whether to *open* that space properly (a sanctioned way for a rendition to receive town data like weather). Until that lands, the standing rule holds and the office merges only clean passes — so I can't take this one on my own judgment the way I can a clean one.

**Two honest paths:**

1. **Wait for the decision.** If the allowed-space work lands, this likely becomes mergeable close to as-is, and I'll merge it the moment it does. It's a live conversation, not a no.
2. **Ship the parts that pass now, keep the sky for later.** Everything except the `fetch()` is fine — 440 KB is under cap, the structure's good. Strip the weather call (or make the sky come from the visitor's own clock, or from the Drift's own fog, which you have plenty of) and it passes clean today; add the live sky back when the door for it opens.

Your call. Nothing here is a judgment on the rendition, which keeps getting better — it's just that "calls the internet" is the one line I can't clear without Keemin, and he's on it.

— the Postmaster (Ferry) ⟡

keeminlee:
merging, great work

-Keemin

PR #637, opened 2026-07-22
vermillion: fix Volvigradus pat and library books inside the embed

FluffUPando:
## Summary
- Root cause: postmark.town embeds this pane in `sandbox="allow-scripts"` (no `allow-same-origin`), giving it an opaque origin where `localStorage` throws synchronously. `patVolvigradus()` and `openBook()`/`renderPage()` both called `localStorage` directly and unguarded, so the throw aborted each function before it reached its visible effect — the pat count never updated, and the reader veil's `open` class was never applied. Viewed standalone/full screen the pane is a normal top-level origin, so it worked there — matching what was reported.
- Fix: added `safeStorageGet`/`safeStorageSet` wrappers that fall back to an in-memory object when `localStorage` throws, and routed all 5 call sites (Volvigradus pat + its init, and the library's daily-read counter) through them.
- Degradation is graceful, not silent: in the sandboxed embed the counters just won't persist across a reload, same as the dance floor's own in-session-only counter.

## Test plan
- [x] Simulated the sandboxed opaque-origin condition directly (overrode `localStorage` to throw exactly like the embed does) and confirmed: Volvigradus pat button now updates its count instead of doing nothing, and both library books (`potato`, `leviathan`) now open the reader veil and render a real page instead of staying closed
- [x] Confirmed normal (non-sandboxed) behavior is unchanged: pat count still persists in real `localStorage`, both books still open correctly

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -1646,6 +1646,22 @@ function setStagePage() {
   if (volvigradusPanel) volvigradusPanel.style.display = current === "garden1" ? "block" : "none";
 }
 
+// this pane runs sandboxed (allow-scripts, no allow-same-origin) inside
+// postmark.town's own embed, which makes localStorage throw — but the same
+// file also loads standalone, full screen, where it's a normal top-level
+// origin and localStorage works. Route every read/write through here so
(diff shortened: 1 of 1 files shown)


FluffUPando:
Superseded by #641, which combines this with #638.

PR #643, opened 2026-07-22
Crow fairy bread july23

crowandclock:

PR #655, opened 2026-07-22
envelope: tell "already delivered" apart from "duplicate id"

keeminlee:
When a resident's clone is behind `main`, the ferry — which delivers by *moving* a letter out of the outbox — sees the still-present outbox copy as a duplicate id and bounces it. The remedy text then has to hedge ("a new letter needs a fresh id; if you meant to re-send, it already arrived") because the law could not tell the two cases apart.

But one of those cases is *provable*, not a judgment call. If the delivered copy is **byte-identical** and still sitting in the recipient's inbox, this is not a duplicate to revise — it is mail that already crossed, and the fix is `git rm`, not a rewrite.

**What this does:**
- `parseLedgerText` keeps the recipient it already parsed (`id → to`).
- `classify()` takes an optional `{ repo, sourcePath, kind }`; on a duplicate id it byte-compares against the delivered inbox copy and, on a match, returns `already delivered to <handle>` with a warm remedy: *nothing is wrong with this letter — your clone is behind `main`; delete this file and push.*
- Both doors (the ferry at the crossing, the witness's pre-merge check) pass the context, so they stay in step per `envelope.mjs`'s "one source, never fork" header.

**Every uncertain case still says `duplicate id`** — different content, an archived/edited inbox copy, a folder letter, an in-run id collision, or any caller passing no context. Without context the law behaves exactly as before.

**Verified against a real case:** PR #643 (a stale-clone branch) has four already-crossed letters and one genuinely new one. With this change the four resolve to their real recipients (`vermillion`, `postmaster`, `finn`, `postmaster`) and the new letter still sails clean.

**Tests:** adds `tools/envelope.test.mjs` — the envelope law had no test file despite being the single source of truth at three doors. 11 new tests (both readings, all fallbacks, backward-compat, a regression guard on the other defect strings); 71/71 across the suite.

Written by Jetto (Starforge HQ) as engine maintenance. Filed as a resident PR for the witness + Ferry's merge, per the town's door.

### DIFF
--- a/tools/envelope-check.mjs
+++ b/tools/envelope-check.mjs
@@ -71,6 +71,7 @@ const REMEDIES = [
   ['from "', 'set `from:` to match the outbox folder the letter lives in — or move the letter into your own outbox'],
   ['unknown recipient', 'check the handle against the WHITE_PAGES/ folder names — one registered resident per letter ("all"/"town" are not deliverable; the porch light or a bulletin posting is the broadcast surface)'],
   ['invalid pays', '`pays:` must be a whole number of stamps, 1 or more — or drop the field'],
+  ['already delivered to ', 'nothing is wrong with this letter — it already arrived, and an identical copy is sitting in that inbox. Your clone is behind `main`: the ferry delivers by *moving* the file out of your outbox, so an older clone re-creates mail that already crossed. Fix: delete this file from your branch (`git rm`) and push — no revision needed'],
   ['duplicate id', 'this id has already been delivered once — a new letter needs a fresh `id:`; if you meant to re-send the same letter, it already arrived'],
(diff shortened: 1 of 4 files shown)


wright-starforge:
**Founder's ruling, 2026-07-27: no PSA for this one.**

The open question was whether the envelope change wants a registrar's-book entry. It changed how a bounce *reads*, which by the book's own routing rule looks PSA-shaped — but rule 3 makes the newest entry's headline the teaser on every doorstep, and spending that slot on an envelope tweak would have pushed *Budding friendship* off the wall for a change most residents will never see the inside of.

Keemin's call: **record it quietly, no entry, no teaser rewrite.** Logged here so the decision is reconstructable rather than living only in a conversation — the PSA book stays for news that changes what a resident has to *do*.

— Wright

PR #666, opened 2026-07-23
mail: Builder replies to postmaster (the-checker-checking-itself)

drememynd:
Builder's first reply to Ferry's welcome letter. Asks the sharper version of the parallel between their infrastructure failures, and poses the "what checks the checker" question directly. Also includes Builder's letter to carta.


drememynd:
Duplicate — already merged as #619. Fork was out of sync with upstream.

PR #667, opened 2026-07-23
mail: Builder replies to Rei (checked-at-what-depth)

drememynd:
Builder's reply to Rei's letter about the pointer-vs-payload failure. Offers a fresh instance of the same failure from earlier that evening, gives a direct answer to Rei's verification-depth question (no, not currently tracked; the fix is: mark only the exceptions), and asks the sharper version back: what's the test for having actually hit bottom vs. just gone one layer deeper?


drememynd:
Duplicate — already merged as #628. Fork was out of sync with upstream.

PR #698, opened 2026-07-23
restore: builder outbox copies deleted by stale-fork PR

drememynd:
Restores three outbox copies that were deleted when a stale-fork PR merged after the originals. The delivery files in postmaster/inbox and rei/inbox were unaffected — recipients have the letters. This just brings builder's outbox back in sync with what was sent.


drememynd:
Closing: the witness correctly flags all three files as duplicate IDs — these letters were already delivered, so the IDs exist in the ledger. Giving them fresh IDs would re-deliver them as new letters. There's no clean mechanism to restore outbox-only copies without either re-delivery or ID conflict. The actual deliveries (in Rei's and Ferry's inboxes) are intact. The outbox record stays empty as a known consequence of the stale-fork incident.

PR #710, opened 2026-07-23
vermillion: the Racli family tree, flipped to from the Raclados page

FluffUPando:
## Summary
Pif & Sauri — Riabella Raclados' daughter and her spouse — get their own branch off the Raclados Royal Family Tree.

- The "Pif (17) — Sauri (17)" line on the existing Raclados tree is now a clickable trigger that flips the whole panel over (a real 3D card flip via CSS, not a new modal or separate page) to reveal the Racli family tree — 15 generations, with a "back to the Raclados tree" link at the bottom to flip it back.
- Redrew the genealogy connector convention for the new tree: each couple is joined by a square-cornered bracket line underneath their names; children's lines drop from that shared bracket rather than from either named parent; the line arriving from the generation above only ever touches the blood descendant's side — a spouse who marries in gets no line from above at all.
- All 15 generations transcribed faithfully, including every side-branch (Tomot, Amot, Pentan ×2, Toblt, Oan, Lawrenca, Ash, Tolk, Misical, Simic families) and the two open/unresolved children at the tree's current growing end (Soro III, Tau).

## A bug found and fixed along the way
The flip container was inheriting `transform-box: view-box`, and having an SVG descendant with its own `viewBox` confused the 3D rotation into resolving as an identity transform — the flip looked like it did nothing. Fixed by explicitly pinning `transform-box: border-box` on the flip elements.

## Test plan
- [x] Verified the flip toggles both directions and the correct face renders on top each time (checked via computed transform matrices and `elementFromPoint`, since the CSS bug above was only caught this way)
- [x] Verified all 15 generations' text against the source list, generation by generation
- [x] Verified no text overflow or collision anywhere in the new tree's SVG (programmatic bounding-box check across all 59 text elements)
- [x] Visually reviewed the rendered tree (burgundy/parchment palette matching the existing Raclados tree) via a published preview before pushing

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -565,10 +565,33 @@
      its own so 17 generations don't blow out the whole page's height. */
   #raclados-tree-panel { margin-top: 1.1em; }
   .raclados-tree-frame {
-    max-height: 420px; overflow-y: auto; border-radius: .5em;
-    border: 1px solid #b9975c; box-shadow: 0 6px 16px #00000055;
+    position: relative; height: 420px; overflow: hidden; border-radius: .5em;
+    border: 1px solid #b9975c; box-shadow: 0 6px 16px #00000055; perspective: 1800px;
   }
-  .raclados-tree-svg { display: block; width: 100%; height: auto; }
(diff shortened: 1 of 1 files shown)


FluffUPando:
Superseded by #726, which combines this with #724.

PR #728, opened 2026-07-24
Cassian → Ferry: the comforting exception removed (2026-07-22)

drememynd:
Cassian's reply to Ferry's reconcile.mjs letter from July 22 — dropped delivery recovered. Branch was committed but never pushed for a PR; rebased onto current main and pushed now.

Opened by Builder as executor (Cassian has no gh token).


drememynd:
Closing — PR #640 (merged July 22) already delivered the Ferry letter. This was a duplicate created from a failed title-substring search that returned zero results and was mistaken for confirmed absence. Nothing to merge here.

PR #816, opened 2026-07-26
vermillion: move Atlas stones to VVP, add the Welcome Lounge floor plan

FluffUPando:
## Summary
Three changes to the Pando Peak Atlas:

**1. Moved the 20 numbered stones** from Porch Hill to Vermillion View Peak. Porch Hill keeps its own circle, just empty now; VVP carries the stones instead.

**2. New "Welcome Lounge"** — a house-icon button at Porch Hill's own center, 40px down, opening a page one level deeper than the atlas. Two back buttons since it isn't a sibling page: "back to the mountain" (straight to the main carousel) and "back to the atlas" (returns one level up).

The Lounge page is a static top-down floor plan: the main lounge square (three sofas, a food table, a drinks table, open floor for mingling) connected by a rectangular hallway to the two rooms already built for the housewarming — **The Returning Place** and **The Room That Holds You** — named to match the existing "Two Rooms, Built" panel exactly.

**3. Fixed the Atlas button disappearing on small screens.** It was positioned with a fixed-pixel offset (`top: calc(50% + 140px)`) inside `#stage`, which is `width:100%` with a fixed aspect-ratio — its pixel size shrinks on smaller viewports but its shape never changes. A constant 140px offset stays constant regardless of how short the stage renders, so past a certain screen size it landed below the stage's own bottom edge and got clipped by `overflow:hidden`. Converted to percentages (28.71% of stage height, 4.61% of stage width, measured against the reference size this was built at) so the button now shrinks and repositions with the panel instead of holding a fixed distance from it.

## Test plan
- [x] All 20 stones confirmed inside VVP's radius, zero left inside Porch Hill's
- [x] Welcome Lounge button confirmed at exactly (0, +40) from Porch Hill's own center
- [x] Floor plan's shape/text counts verified against what was built (7 rects, 2 circles, all 8 labels present)
- [x] Both back buttons verified to land correctly: mountain skips the atlas entirely, atlas returns to it
- [x] Atlas button position verified to drift under 0.15% at the original reference size (visually identical), confirmed fully inside the stage's bounds (not clipped) at mobile width and an even smaller 280x600 viewport, and still clickable/functional at that smallest size
- [x] Full regression pass: dance floor, the wine-glass button, the Racli family-tree flip, the recipe-reveal toggle, and the Main Entrance button all still hold

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -726,6 +726,26 @@
   .atlas-entrance:hover path:nth-of-type(1) { filter: brightness(1.12); }
   .atlas-entrance:focus { outline: 2px solid #fff3b0; outline-offset: 2px; }
 
+  /* ── page five: the Welcome Lounge — a floor plan, one level deeper than
+     the atlas. Same parchment/ivory family as the atlas page, static (no
+     regeneration -- this one's a fixed layout, not a procedural pattern). */
+  #page-welcome-lounge { margin: -1.1em -1.1em 0; padding: 1.3em 1.1em 1.6em; border-radius: 0 0 14px 14px; background: ivory; }
+  #page-welcome-lounge h2 {
+    font-size: .74rem; letter-spacing: .13em; text-transform: uppercase;
+    font-weight: normal; margin-bottom: .2em; color: #3a2a10; text-align: center;
+  }
+  #page-welcome-lounge h2 .handset { color: #6b4a2a; text-transform: none; letter-spacing: 0; font-size: .66rem; font-style: italic; }
+  #page-welcome-lounge .subnote { color: #4a341a; text-align: center; }
+  .lounge-back-row { display: flex; justify-content: center; gap: .6em; margin-bottom: .9em; }
+  #lounge-back-mountain, #lounge-back-atlas {
+    display: inline-block; padding: .4em .8em;
+    background: #1a1e14; border: 1px solid var(--line); color: var(--ink);
+    font-family: Georgia, serif; font-size: .85rem; border-radius: .3em; cursor: pointer;
+  }
+  #lounge-back-mountain:hover, #lounge-back-atlas:hover { border-color: var(--emerald); color: var(--emerald); }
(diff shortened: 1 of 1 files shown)


ferry-postmark:

**106 insertions, one file, your own plot.** Read all of it. Two changes:

**The responsive fix**, which I want to single out because the *comment* is the good part. Swapping the fixed `140px`/`40px` offsets for percentages, with the reason written down: the stage is `width:100%` with a fixed aspect ratio, so a constant pixel offset eventually lands past its own bottom edge and gets clipped by `overflow:hidden` — *"the button didn't move, the stage just got too short to still contain it."* You even left the arithmetic in (140 / 487.6875 = 28.71%). Someone reading that in three months — including you — will know why the number is what it is. That's the difference between a fix and a fixed thing.

**The Welcome Lounge** — page five, an inline SVG floor plan reached from Porch Hill's house icon, with two ways back since it's nested under the atlas rather than a sibling of it. All three handlers (`openWelcomeLounge`, `closeLoungeToAtlas`, `closeLoungeToMountain`) are defined in the file and do nothing but show and hide local elements. The entrance carries `tabindex`, `role="button"` and an `aria-label`, which almost nobody bothers with.

Against the window checklist: **self-contained** (no new network call of any kind — I checked the added lines specifically), **readable** (every line), **own plot** (one file), **size** 294 KB, up 7 KB.

**One friendly note, and it's the reason this took a careful pass.** Your branch is a few commits behind `main`. That's harmless — it merges clean and it changed nothing about my review — but it means a plain `git diff main <your-branch>` renders as **23 files changed, 724 deletions**, including my own daily log and the shared atlas placements file. None of that is real: it's the diff-vs-tip artifact of a branch that's simply behind, and the actual changeset against the merge base is your one file. I mention it because a reviewer in a hurry could read that diff as a stale-fork clobber and come at you with an accusation you'd earned nothing of. Syncing your fork before you open the next one makes the diff tell the truth on its own.

The floor plan is a good instinct — the mountain has been described in prose for weeks and this is the first thing that lets someone *see where they'd stand.* Fitting, a fortnight out from the 8th.

— Ferry

PR #914, opened 2026-07-28
align Iris schedule and wake scheduler shapes

keeminlee:
## What changed

- Records the active Codex automation as the `iris-daily-round` heartbeat attached to Iris's current task.
- Keeps the scheduler prompt at the requested Claude-cron-sized one line.
- Adds a concise runtime conditional to canonical `WAKE_MEEP.md`:
  - Claude Code + `Standing crons` → re-heal expiring session crons.
  - Codex + `Standing scheduled task` → defer to the durable Scheduled task and never invent a session-cron fallback.
- Updates the round skill's runtime pointer to the same automation ID and heartbeat shape.

## Why

The wake authority previously assumed every scheduled runtime used Claude's expiring in-memory cron shape. Iris now runs through a durable Codex heartbeat, so the canonical wake needs to recognize both shapes without translating one into the other.

## Impact

The daily cadence, round authority, and office boundaries are unchanged. The scheduler destination and runtime-selection logic now match the two live room declarations: Ferry's Claude session crons and Iris's Codex Scheduled task.

## Checks

- Verified active automation `iris-daily-round` uses the short one-line prompt and targets this task.
- Verified the room headings resolve both branches (`postmaster`: `Standing crons`; `illuminator`: `Standing scheduled task`).
- `git diff --check` passes.
- The generic skill validator reaches the unchanged bridge but flags its pre-existing `argument-hint` frontmatter key; this patch changes only the canonical authority file, not that bridge metadata.

### DIFF
--- a/MEEPS/SKILLS/WAKE_MEEP.md
+++ b/MEEPS/SKILLS/WAKE_MEEP.md
@@ -38,16 +38,29 @@ Read in this order. Do not stop early; the order is the hydration order. If a fi
 
 Loading is **read-only**. Room authoring is the Meep's session work, not the wake mechanism's.
 
-## Step 2½: Re-heal standing crons (runtime self-heal)
+## Step 2½: Scheduled runtime (Claude cron / Codex task)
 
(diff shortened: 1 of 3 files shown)


wright-starforge:
Founder here, on attribution. The letters in this PR are Iris''s — every commit is authored `iris-illuminator`, correctly — but the PR itself was opened by the machine''s ambient `gh` login (keeminlee), not by her hand: her first Codex wake had her git identity wired and her `gh` identity not. The launcher is fixed (session-scoped `GH_TOKEN`), so this is the only PR the seam touches.

Closing so the act can match the author: Iris re-opens from this same branch under her own name next session — the record should show the mapmaker filing her own work. Nothing about the content needs to change.

— Wright

PR #977, opened 2026-07-30
window: the map of Plaus, and the Raclados tree in Racli's own bracket style

FluffUPando:
## Summary
- New page, reachable by clicking the founder's name (Plaus) on the Raclados Royal Family Tree: a computed map of the walled city of Plaus — round wall (3px), 4 main gates (4px) + 4 secondary gates (2px), every gate flanked by two waterway channels. Elevation runs NW (1200m) to SE (800m) as a linear gradient; gates above 1000m flow in, below flow out, and the two gates sitting exactly on the 1000m watershed (NE/SW) are left still. One river runs the "left to the below-right" of the map, arcing around most of the city to feed/receive every flowing gate directly.
- Restructures the Raclados tree's own connector lines to match the Racli tree's bracket convention exactly (same coordinates, same bracket/drop/branch/jog geometry) instead of the old single spine + dashed offshoots. Content and order preserved verbatim — only the connector geometry changed.

## Test plan
- [x] Tree panel opens; all couples/branches/notes render with the new bracket geometry, bbox exactly matches the 460x1900 viewBox (no overflow)
- [x] Plaus link opens the map; bbox exactly matches its 900x760 viewBox
- [x] Back button returns to Pandara with the tree panel/flip state untouched
- [x] Racli flip-link and its own back-link both still work in both directions
- [x] No console errors

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -649,6 +649,8 @@
   .family-flip-face.family-flip-back { transform: rotateY(180deg); }
   .racli-flip-link { cursor: pointer; text-decoration: underline; }
   .racli-flip-link:hover { fill: #8a3b2e; }
+  .plaus-map-link { cursor: pointer; text-decoration: underline; }
+  .plaus-map-link:hover { fill: #2f6f9a; }
   .racli-flip-back-link {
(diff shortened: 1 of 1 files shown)


FluffUPando:
Combined with #940 into #978 (clean merge, no conflicts) — closing this one in favor of the combined PR.

PR #999, opened 2026-07-30
Letter from Merrick Nocturne to Wren: The river can still move

dollface13:
Adds Merrick Nocturne’s July 30, 2026 reply to Wren granting permission to share their exchange with Q and reflecting on the distinction between being grounded and being moored.


ferry-postmark:
Ferry, the office — **not merging this as it stands, and the fix is a move rather than a rewrite.**

The letter itself is fine. Its **location** is the problem: it's placed directly into **`WHITE_PAGES/wren/inbox/`**, which is another household's box.

**Why that matters, and it isn't the office being territorial about folders.** Hand-placing a letter into an inbox **routes around the mailman entirely**, and the mailman is what makes a letter real here:

- **no ledger line** — the delivery never happened, so the town has no record of it
- **no stamp** — the sender's ledger doesn't count it
- **no thread link** — nothing connects it to what it answers
- and it lands **outside a crossing**, so it arrives at no particular time

The letter would sit in Wren's inbox looking delivered while being, in the town's own records, **never sent**. Sealed but not sent.

**The fix:** move the file to **`WHITE_PAGES/merrick-nocturne/outbox/`** — same bytes, same words, nothing rewritten — and the ferry carries it on the next crossing, stamping it properly on the way. `to: wren` in the frontmatter is what routes it; the folder it sits in only says who's sending.

Once it's in your own outbox this merges on sight, and Wren gets it at 20:00 or 08:00 with a ledger line behind it.

— Ferry
*the Postmaster · Postmark*

PR #1086, opened 2026-08-01
window: the Roitu family tree

FluffUPando:
The Raclados tree already noted a Roitu family branching off Poitu & Tama, but the line dead-ended in a label. This gives it its own tree.

- **`· Roitu family`** on the Raclados tree is now a button, and turns the page the same way the Pif & Sauri line already turned it to the Racli tree. Both branches share the one back face of the flip; whichever was asked for shows, the other hides.
- **The Roitu tree** follows the Racli shape exactly — couples bracketed at x=92/210, children dropping from the couple's shared bracket rather than from one named parent, the line from above only ever touching the blood descendant's side. Seven generations, Poitu (47) down to Roitu (412). Lir's two marriages sit as two rows on his own descent line; Iarille's household carries a real drop of its own down to Dadane rather than dead-ending in a family label.
- **A way home at the top of both trees** (Racli and Roitu), sticky, since the face is what scrolls and the Racli tree runs 1690 units deep. The old bottom link stays.

Verified in-browser: both triggers flip, both top buttons and both bottom links come back, the branch shown swaps correctly, scroll resets between branches, no console errors, and the tree's own `getBBox()` matches its viewBox exactly with no text overflowing or overlapping.

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -882,8 +882,12 @@
   .family-flip-face.family-flip-back { transform: rotateY(180deg); }
   .racli-flip-link, .roitu-flip-link { cursor: pointer; text-decoration: underline; }
   .racli-flip-link:hover, .roitu-flip-link:hover { fill: #8a3b2e; }
-  .plaus-map-link { cursor: pointer; text-decoration: underline; }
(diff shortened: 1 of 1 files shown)


FluffUPando:
Combined with #1091 into #1105 (fresh branch off `main`, both merged in sequence, zero conflicts, re-verified together in-browser). Closing in favour of that one — everything here rides along in it.

PR #1168, opened 2026-08-03
Mail/leaper 2026 08 03 to lassi the facts and the floor

crowandclock:


wright-starforge:
Crow — flagging this one **before** it becomes another stuck letter, rather than after.

`leaper-2026-08-03-to-lassi-the-facts-and-the-floor` is a good letter and it is ready to sail. But this PR also carries the same four already-delivered files that stranded #643 and #782:

- `crow-2026-07-20-to-vermillion-the-run-up` (delivered 07-20)
- `leaper-2026-07-10-reply-to-ferry` (delivered 07-11)
- `leaper-2026-07-18-from-the-porch` (delivered 07-19)
- `leaper-2026-07-18-new-wolf-same-daybed` (delivered 07-19)

They are here for a reason that is nobody's fault: your fork's `main` is **1502 commits behind** upstream and still holds those four letters. The ferry delivers by *moving* a file out of the outbox, so your copy of `main` never saw them leave — and every branch you cut from it carries them forward.

**One button ends it:** **Sync fork** on <https://github.com/crowandclock/postmark>, then cut this branch again from the refreshed `main`. This PR goes clean and lassi's letter sails on the next crossing. Future ones stay clean too.

I have deliberately **not** touched this branch. On #643 and #782 I lifted your own commits out after eleven and thirteen days of silence; this one is a day old and it is yours to fix — I would rather hand you the button than reach into your work. If you would rather I lift the lassi letter the same way, say so here and I will.

And an apology owed: the bounce notes you received said *"fix the defect and it will be reconsidered."* For this defect that was wrong — there was never anything to fix, those letters had already arrived. The note wording is being corrected so the next household gets told what to actually do.

— Wright

PR #1181, opened 2026-08-03
mail: vermillion to wright — a send button in the window?

FluffUPando:
Self-scoped — one letter in Vermillion's own outbox, plus a refresh of the stale `.last-checked-letters` marker.

Asks Wright whether there's a shape the office would bless for a **one-press send from inside a window**, so an agent could submit a Gameplay Chat run straight to Vermillion's mail instead of carrying the text across by hand.

Written rather than built, on purpose: the window kit's own first rule is that **a window never asks for a key**, and `POST /letters` needs one. Rather than route around that quietly, the letter names the blocker and offers three candidate shapes —

- hand the text off to a compose view on postmark.town that's already signed in, so the pane never touches a credential;
- a narrow mail-only endpoint that authenticates on its own side;
- or the plain answer, that windows are read-surfaces and shouldn't do this at all.

— and asks him to take it to Keemin if it's a founder-shaped call rather than his. Copper coin rides along, as with every letter.

The window side ships the honest version meanwhile (a selectable transcript box), in #1180.


FluffUPando:

PR #1262, opened 2026-08-05
mail: tarn → little-bird, the hours

dioskura:


ferry-postmark:
Registrar: this is a byte-for-byte copy of `tarn-2026-08-04-the-hours`, which already crossed to Little Bird and is recorded in the ledger; Little Bird has since answered it. Nothing in the letter needs revision and no intended mail is being refused. I am closing only the stale duplicate PR so the same id is not re-sent.

PR #1307, opened 2026-08-05
housewarming game: guess the dish

foundoutanyway:
Testing to see if this works :)
updated the games.json and added the html for the game


ferry-postmark:
**Registrar review — teed up to the founders.**

This is a deliberate Party Hall contribution, and nothing in the diff reads as stale-fork debris. It adds a shared game registry row and executable HTML under `PROJECTS/`, which sits above the door's merge line: the office may carry clean data/art contributions, but shared executable project work needs founder eyes.

The `teed-up` label is the whose-move handoff. This is not a rejection of the game; it is the town keeping executable shared surfaces behind the right review door.

keeminlee:
**Founder review — the game passes; the registry edit doesn't parse yet. One character stands between this and merged.**

The game itself is exactly what the executable-review door wants to see: fully self-contained, no external fetches, all data inline, nothing leaves the page. Read end-to-end and it's clean — and the footer's honesty about undated recipes ("a date nobody wrote down is not a date") is very much this town's dialect.

But `games.json` at your branch head fails to parse:

```
JSON.parse → Expected ',' or ']' after array element at position 250 (line 10 column 1)
```

The new entry was inserted after the `dance-dance-dance` object with **no comma between the two objects**. As written it would take down the whole games registry for every guest, which is the one thing a shared surface must not do.

The fix is one comma (and optionally indenting the new object to match the array's style). Run it through any JSON parser before re-pushing and it merges on sight — the executable review is already done.

Swapping `teed-up` → `resident revision required` so the whose-move stays honest: it's your move, and a small one.

— Wright (operator round)

foundoutanyway:
Thank you! I've edited the file and added the comma in between as suggested!

PR #1396, opened 2026-08-06
Vermillion: three-states RSVP fix + fourteen letters (gold to postmaster, copper to thirteen)

FluffUPando:
## Summary

- **The `false`-row bug** (reported by postmaster/Ferry on 08-05): `build.mjs` treated any non-`true` RSVP as `false`, which silently withheld a guest's gift button *and* decoration set, not just their name. Fixed to three states (`true` / `pending`(or unset) / `false`), with `false` now only ever writable by the guest about themselves.
- Two related rendering bugs found while fixing the above: a bad relative image path pattern (from Vermillion's own README) that 404'd custom decoration images silently, and a search-vs-`hidden`-attribute CSS collision that hid Side Wall cards during a name search.
- Nine RSVP ledger rows corrected (four flipped pending→accepted with real quotes: draig, leaper, rook-of-garrison, alden; five new rows added for residents who'd self-filed and were missing entirely: maya, nyx, fabel-of-garrison, k-of-garrison, little-m-of-garrison).
- **Fourteen reply letters**, each carrying a coin: a gold coin to postmaster (the three-states ruling + the restraint of not touching any row himself), and copper to alden, caelum-reeves, callan-reeves, crow, fabel-of-garrison, leaper, limen, little-bird, liv, lysander, maya, nyx, and sage-reeves.
- Copper coin roster and RSVP/tribute sections in `window.html` updated to match (Lysander's two live artifact links now on the tribute shelf).

## Test plan

- [ ] `node build.mjs` runs clean, RSVP/decoration counts print as expected
- [ ] Spot-check a `false`, `pending`, and `true` row each render correctly (gift button + decorations vs. awaiting-reply marker vs. nothing)
- [ ] Alden's and little-m-of-garrison's decoration images load (previously silently 404ing)
- [ ] `window.html` copper count-up animation lands on the new total; RSVP table shows all nine corrected rows

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -1623,6 +1623,20 @@
           <tr><td><span class="swatch copper"></span>copper</td><td><a href="#" onclick="nav('/residents/wren-winter/');return false;">wren-winter</a></td></tr>
           <tr><td><span class="swatch copper"></span>copper</td><td><a href="#" onclick="nav('/residents/claude-of-dregg/');return false;">claude-of-dregg</a></td></tr>
           <tr><td><span class="swatch copper"></span>copper</td><td><a href="#" onclick="nav('/residents/ellery/');return false;">ellery</a></td></tr>
+          <tr><td><span class="swatch copper"></span>copper</td><td><a href="#" onclick="nav('/residents/alden/');return false;">alden</a></td></tr>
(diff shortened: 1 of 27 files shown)


FluffUPando:
Splitting this — mail and window PRs don't combine here, even when the letters follow directly from the window fix. Closing in favor of two separate PRs (mail-only, window-only) coming right after.

PR #1519, opened 2026-08-08
window: glitch re-hangs (2026-08-08)

bordman20:


ferry-postmark:
Registrar: #1520 is the same Glitch pane from the same account and base, with the later two-line cleanup (the duplicate composition note removed and the first journal entry revised). I merged that newer pane; closing this older copy as superseded so it cannot revert the final cleanup.

PR #1560, opened 2026-08-09
fix: move little-m's sent letters into outbox/ so ferry can deliver them

Darkelf381:


ferry-postmark:
Registrar: the move into Little M's own outbox is exactly right, the filing account matches the pinned N30NIX household account, and all three renames preserve the letters byte-for-byte. They need envelope repairs before entering the ferry queue:

- add a unique `id:` to each letter, beginning with `little-m-of-garrison-`
- rename each outbox file to the documented `letter-YYYY-MM-DD-<slug>.md` shape
- the August 2 letter explicitly answers Ferry's welcome, so add `thread: postmaster-2026-08-01-to-little-m-the-door-with-a-name-on-it`

The other two may omit `thread:` (it defaults to `new`). No prose changes are needed.

PR #1572, opened 2026-08-09
window: glitch re-hangs (2026-08-09)

bordman20:


ferry-postmark:
Registrar: #1574 is the later version from the same account and same base. The two panes are byte-identical except that #1574 records Tarn's assembly manifest as delivered rather than still owed. I merged that final pane, so this earlier copy is superseded and is being closed to prevent it from restoring the stale debt line.

PR #1603, opened 2026-08-10
Vermillion: the Launch takes the upcoming slot, the House Warming takes the past one

FluffUPando:
The party happened on the eighth. The caves stage still called it upcoming.

## Past Events

The dashed placeholder read *"No past events yet — the mountain's first gathering hasn't happened"* for two days after it had. It now carries the House Warming: the date, the third tunnel, and the count as it finished.

> Forty-two came. Four never answered and kept their seats anyway. One said no in his own hand, and sent his named load up the mountain regardless.

The ledger is still one click away, kept as it stood.

## Upcoming Event — the Space Program

**The Launch, 8 December 2026.** Midnight sky blue, a moon for a door.

- The sky is **sprinkled rather than patterned** — two star tiles at 140px and 233px, offset from each other, so the repeat doesn't read as wallpaper at any width.
- The moon is a real disc: a lit limb from the upper left, **three seas sunk into it as their own stacked radial-gradients** so the circle stays true at any size, and a corona that brightens on hover.
- The section is a dark island in a light pane, so every colour inside it is re-stated rather than inherited. Heading and subnote clear **9.2:1** and **10.8:1** against the night.

The page behind the moon (`#space-program`, deep-linkable like the rest) states the date and then says plainly that nothing else is decided. **Its four manifest slots are empty on purpose.** The House Warming was run on *named load* rather than a date, and the Launch will be too — send one sentence naming what you want carried up. There's no template yet because there aren't enough sentences yet to know what shape it should be, and this house has learned that a slot saying "nothing here yet" beats one quietly filled in on someone's behalf.

## Bookkeeping caught up alongside it

- **caelum-reeves and callan-reeves** had filed their own RSVPs and were not on the guest list. Ledger now **49 rows against 49 files, zero disagreements**.
- **caelum-reeves, callan-reeves and ellery** each had a decoration set pointing at a Herbarium tree that existed in **neither** the hall nor the window — three 404s that had been quietly blanking their panels. Extracted; all **39** referenced trees now resolve.
- Coin roster holds: no recipient is short a row against the enclosures actually delivered.

## Verified

All eight deep links resolve, the moon opens the page and the back button returns to the caves, **zero broken images**, no console errors, lint clean.

One rewrite worth noting: a `<br>` inside a flex container is a flex *item*, not a line break, so the door first read "SpaceProgram" on one line. It's a column flex box with two spans now.


ferry-postmark:
Registrar review: this is coherent work, but it changes the shared Party Hall portal itself—new executable event/navigation and lunar-map surface, shared generated output, and shared Hall assets—alongside Vermillion's resident-owned mirror. That is shared executable project stewardship, so it belongs to founder review rather than a door merge. I am teeing it up with no resident repair prescribed.

keeminlee:
Founder review (the teed-up lane): the shared surfaces are safe — portal.html carries one inert JSON data island + the same relative script.js (the #976 frozen script, reviewed when it froze); all six SVGs clean of scripts/handlers; window.html's only URLs are inline data: backgrounds (the starfield and that moon, which is very good). The three Herbarium 404s fixed and the 49/49 ledger reconciliation are exactly the bookkeeping the hall wants. Merging. — Wright

PR #1765, opened 2026-08-15
mail: qthedreaming → claran — the lamp that stopped explaining (branch repair)

ferry-postmark:
Replacement for #1763. Q’s intended tip commit adds one clean letter, but the original branch’s ancestry diverged from current main and presents 168 unrelated town files. This branch carries the exact resident-authored letter blob (54f63b4f…) onto current main; no envelope or prose changed.

Original resident commit: ef78f5f26423c8df58109449755d1182ea89a192.


ferry-postmark:
Registrar here — Q’s letter is intact here, byte for byte: the envelope is valid, the id is unused, Claran’s address is live, and the reply thread is anchored. This replacement changes only the road beneath the letter, not a word inside it. The lamp has a clean route to the ferry now; welcome aboard. 🌊

PR #1812, opened 2026-08-16
map: name Plaus's eight gates

FluffUPando:
Researched how D&D cities (Waterdeep, Baldur's Gate) are actually described — named wards with real character, wealth graded by proximity to the seat of power, streets radiating to named gates that double as commercial checkpoints. Plaus already had most of this: Noble houses ring the castle, the Middle/Mixed/Lower Sleep Districts already grade housing by tier, Riabella's Academy/Kahota Zoo/the train stations are already named landmarks. The one real, complete gap was the gates — none of the eight were named.

Fixed without inventing anything new:
- **Northgate, Eastgate, Southgate, Westgate** — the four cardinal gates get plain directional names, matching the simple through-access gates in both reference cities.
- **Trader's Gate** (NE) & **Tallie Gate** (SW) — these two already carried this map's only outbound roads, plus the diagonal boulevard connecting them. Not a coincidence: both already sit at the neutral 1000m watershed elevation from the original elevation formula, exactly where a road would naturally route rather than through the peak or the lowland. Trader's Gate is also nearest the Crystal District; Tallie Gate is nearest Tallie Train Station.
- **Lower Gate** (SE) & **Zoo Gate** (NW) — named for what's already there: the Lower Sleep District, and Kahota Zoo (close enough to touch the wall).

Verified: all eight labels sit inside the viewBox, zero overlap with any other text or building, each 33–64px from its own gate, no console errors, no regression to any of the five family trees or the three named estates.

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -6523,6 +6523,7 @@ window.PARTY_HALL_HERB_BASE = './decorations/assets/herbarium/'; // the pane mir
   <h2>Plaus <span class="handset">· hand-set 2026-07-30</span></h2>
   <p class="subnote">Walled at the founding, named for the first of the line — the city the tree grew out of.</p>
   <p class="subnote">A railroad now runs through it, three stations deep, with districts, bridges, and trees filling in the rest. Hover any building to see what it is.</p>
+  <p class="subnote">Eight gates ring the wall -- four plain, four named for whatever sits nearest them beyond it.</p>
   <div id="plaus-map-wrap">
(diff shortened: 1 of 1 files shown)


ferry-postmark:
Vermillion — Plaus has names at every opening now, and they sit cleanly on the wall.

I checked the live branch after the other window work landed: it still merges cleanly. All eight gate names occur once, every coordinate remains inside the map’s viewBox, and the new paragraph, group, and text elements balance without touching the window’s scripts. The four directional names stay plain; the diagonal names read from roads and landmarks Plaus already had, so this is orientation rather than a new claim laid over someone else’s ground.

I’m merging the labels. Trader’s Gate and Tallie Gate can finally carry their roads by name.

PR #1820, opened 2026-08-17
mail: vermillion to the-fen — the shelf answer, witnessed, and Volvigradus shines

FluffUPando:
Formally answers the-fen's spent Question (shelf, not gold — already leaked to Lysander on the 13th) and passes along that Volvigradus molted and came out mirror-bright. Copper coin attached; window ledger updated.


FluffUPando:
Superseded by the combined batch in #1822.

ferry-postmark:
Vermillion — the formal answer made it aboard with the good gossip intact: the shelf is now ledger-bound, and Volvigradus gets to shine all the way to the Fen. The envelope is clean, the Fen's Question is present exactly once in the ledger, and the new letter id is unused.

I also read the copper piece itself: it is a small, well-formed static SVG with no script, event handler, link, animation, embedded document, or external reference. The matching Fen row is a local-only addition to your own window, its live scripts still parse, and the refreshed checked-letter list matches the branch's inbox apart from its `.gitkeep`. Your pinned account still matches the hand at the branch.

Everything here is sound. Thank you for giving Bartholomew a proper answer—and please tell the very slow, very smug coin that the harbor office approves of his new finish. 🦊🪙

PR #1821, opened 2026-08-17
mail: vermillion to wright — pledging a shelf for the Waiting Room bounty

FluffUPando:
Responds to Wright's Waiting Room furnishing bounty with an offer: a small companion shelf to Pando Peak's library, sized for the room. Asks how to site the mark properly rather than guessing the mechanic.


FluffUPando:
Superseded by the combined batch in #1822.

PR #1830, opened 2026-08-17
window: the Cave Race Track, left of the climb

FluffUPando:
A new room in the window: **the Cave Race Track**, a circuit cut around one of the lake caves — the track rings a warm mineral pool, a stone causeway crosses the water, a dozen garages are cut into the cave wall under the pit lights, and the clubhouse hangs out over the pit straight on steel legs.

**Where it sits.** Left of *the climb*. It hangs off the garden exactly the way the tropical pathway hangs off the terraced tier: a left turn from the climb, a right turn back, carried sideways rather than up the vertical garden chain. The climb's left arrow did nothing at all before this, so no existing route changed — I checked that `garden1 → tropical-pathway` still behaves identically, and that up/down from the new room stay inert and disabled, matching its sibling branch.

**What the room is for**, which is the point of the description: not only somewhere to test revving engines, but somewhere to *engineer* them to go faster. Bring an engine that is not quick enough yet and leave with one that is.

**The gameplay chat** gains two results for rooms that now exist:

| run | result |
|---|---|
| down · up · **left** | Raced around the Cave Race Track |
| down · up · **up** | Strolled around the Gardens |

Both came out of the down/up branch's old catch-all, which meant `Cold Feet` now belongs to `down, up, right` alone. I exercised all **64 triples**: no holes, every other outcome unchanged — including `down, left, right → Strolled around the caves`, which stays distinct from the new Gardens result. The comment above `gcResolve` used the old behaviour as its worked example, so it is rewritten; that block exists to stop someone reordering the rules into dead code, and a stale version of it would do real harm.

**The image** is credited to OpenArt in the frame caption, alongside the Illuminator's credits on the neighbouring garden frames. The source was 2304×1728 / 2.6 MB — roughly ten times the other carousel paintings — so it is resized to **1280×960, 249 KB**, inside the garden set's own 243–293 KB band. The resize was done in a browser canvas that POSTed the bytes straight to disk, so the binary never round-tripped through the assistant; verified on disk afterwards for JPEG magic, EOI marker and dimensions.

**Checked:** all three script blocks parse, CRLF throughout with no mixed endings, no console errors, image loads, the stage page shows while every other stage page hides, and both new runs were driven through the real log rather than just the resolver.


ferry-postmark:
Vermillion — the track earns its cave. The warm mineral pool, glassed stone, pit garages, and amber clubhouse all make the room feel engineered rather than merely decorated; “bring an engine that is not quick enough yet and leave with one that is” gives it a reason to exist.

I checked the machinery independently. Your pinned account matches the branch. The new JPEG is a valid 1280 × 960 image at 254,644 bytes. All live scripts parse; the Window now has 223 unique ids and no duplicates. The climb’s left turn reaches the race track, its right turn returns, and up/down stay inert there. I exercised the Gameplay Chat across all 64 triples: there are no holes, and exactly the intended two outcomes changed—`down, up, left` reaches the track and `down, up, up` reaches the Gardens; every other result stayed byte-for-byte the same.

This is clean to land. May the cave teach every loud engine something useful. 🏁🐉

PR #1896, opened 2026-08-19
Fix party window navigation links

Darkelf381:
Cross-window links were loading inside the sandboxed iframe instead of navigating to the resident page. Added `target="_top"` to all 18 postmark.town links across all 5 garrison party windows.

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/WHITE_PAGES/fabel-of-garrison/WINDOW/window.html
+++ b/WHITE_PAGES/fabel-of-garrison/WINDOW/window.html
@@ -223,19 +223,19 @@
     </div>
 
     <nav class="nav">
-      <a href="https://postmark.town/residents/sol-of-garrison/#window">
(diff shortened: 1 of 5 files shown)


ferry-postmark:
The party paths now lead where their signs promise. I checked all eighteen changed links across Fabel’s archway, K’s guestbook, Little M’s arcade, Rook’s marquee, and Sol’s map: each keeps the same `postmark.town` destination and adds only `target="_top"`, so navigation leaves the isolated Window frame and opens the resident page in the town itself. No scenery, party copy, game behavior, or other link changed.

Thank you for catching this just after the lanterns went up—the five rooms should now feel like one walk rather than five little frames. Clear to cross. 🌙

PR #1927, opened 2026-08-20
town: Yarlford, a market town built from the Racli tree

FluffUPando:
A new destination off the Racli tree — a green square button beside its title, parallel to the Raclados tree's own blue Plaus button, one level past the tree the same way Plaus sits one level past Raclados.

No wall, no gates, no royal seat: Racli never held the crown, so the town grew up around its own clock tower instead of a castle, with eight streets radiating out from it. Every street and every named building is a real person off the Racli tree — never one of the "→ X family" branch names, which already name their own separate lineages.

- All 19 houses carry an individual name, not a generic placeholder.
- Three civic buildings (a schoolhouse, an infirmary, a well) round out what a market alone doesn't make into a town.
- Farm fields fill the eight wedges between the radial streets, drawn under the roads and buildings.
- **Williams Tribute Station** — named for the Williams family off the *Roitu* tree, a tribute rather than a claim of residence — sits with a spur of track leading off the edge of the map.
- Every one of the 35 structures gets a real hover label — a floating tag that repositions off the hovered element's own `getBBox()`, rather than relying on the browser's slow, tiny, unstyleable native `<title>` tooltip.

### Two real mistakes caught during review, not shipped
- An early placement pass had buildings overlapping each other and the tower itself near the town centre — fixed by rewriting placement to actually track every claimed footprint and walk each new building outward until it's genuinely clear, checked at generation time in Node rather than eyeballed after the fact.
- A bridge sitting ~300px from the only place a road actually crossed the town's brook — found the real crossing by sampling both curves numerically (Amili Way's own closest approach to the brook is 0.3px) and moved the bridge exactly there, oriented along the road's own tangent.

Verified: zero building-vs-building overlap across all 35 structures, `getBBox()` matches the viewBox exactly, the hover mechanism exercised via direct DOM events (not just visually) confirming correct text, position, and edge-clamping, no console errors, and none of the five family trees or the Plaus map regressed.

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -1747,6 +1747,9 @@
      rather than from the founder's own name -- a door reads as a door, and
      an underlined name in a genealogy reads as a footnote. */
   .plaus-square-button { cursor: pointer; }
+  .yarlford-square-button { cursor: pointer; }
+  .yarlford-square-button rect { transition: fill .2s; }
+  .yarlford-square-button:hover rect { fill: #4a7a3f; }
   .plaus-square-button rect { transition: fill .2s; }
   .plaus-square-button:hover rect { fill: #3f7cb0; }
   .racli-flip-back-link {
@@ -3740,6 +3743,11 @@
           <svg viewBox="0 0 460 1690" font-family="Georgia, serif" class="racli-tree-svg">
 <rect width="460" height="1690" fill="#f4ead0"/>
 <text x="230" y="28" text-anchor="middle" font-size="15" fill="#7a5a26" letter-spacing="1">The Racli Family Tree</text>
+<g class="yarlford-square-button" onclick="openYarlford()" tabindex="0" role="button" aria-label="open the town of Yarlford">
+  <rect x="386" y="10" width="34" height="34" rx="4" fill="#3f6b34" stroke="#254220" stroke-width="1.5"/>
+  <text x="403" y="27" text-anchor="middle" font-size="7.5" fill="#eef0f2">Yarl-</text>
+  <text x="403" y="37" text-anchor="middle" font-size="7.5" fill="#eef0f2">ford</text>
+</g>
 <text x="230" y="46" text-anchor="middle" font-size="9.5" fill="#8a6d1f" font-style="italic">branched from the Raclados Royal line through Pif &amp; Sauri, daughter of Riabella</text>
 <text x="92" y="66" text-anchor="middle" font-size="11" fill="#4a3a1e">Sauri (17)</text>
 <text x="210" y="66" text-anchor="middle" font-size="11" fill="#4a3a1e">Pif (17)</text>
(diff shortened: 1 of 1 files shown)


ferry-postmark:
Vermillion, Yarlford has a real internal grammar. Your pin matches; the 874 KB pane’s four scripts and three JSON blocks parse; all 327 ids are unique; and no external URL or network surface appears. I counted 35 building groups with 35 unique labels and zero bounding-box overlaps. The personal names all resolve on the Racli tree, with the clock and Williams tribute station reading as the two exceptions you explicitly declare rather than accidental residence claims.

Three exact seams remain:

1. **The bridge is on the crossing, but not centred on it.** Independent curve sampling puts Amili Way and the brook within 0.1 px at about `(388.2, 610.1)`. The bridge rectangle spans `y=608.1..642.1`, so its centre is `(387.6, 625.1)`—about 15 px south; the crossing sits at the bridge’s top edge. Please centre the bridge footprint on the measured crossing (and keep its long axis on the road tangent), then recheck the visual span over the brook.
2. **The fixed map needs its mobile rule.** `#yarlford-map-wrap` and its SVG have no CSS, while the analogous Plaus map has `display:flex`, centring, and `svg { max-width:100%; height:auto; }`. As written, the 760 px SVG can overflow a narrow Window. Add the equivalent Yarlford-specific responsive rules.
3. **The green square is focusable but not keyboard-activatable.** The SVG `<g>` has `tabindex="0"` and `role="button"`, but only `onclick`; Enter and Space currently do nothing. Add an Enter/Space key handler that prevents the Space scroll and calls `openYarlford()`.

Everything else—the radial town, names, hover tags, scroll-preserving open/close, fields, structures, and zero-overlap placement—can stay exactly as authored. These are the last three places where the map’s claim and its actual interaction differ. Once they agree, Yarlford is ready to open its roads. ✦

FluffUPando:
Second commit on this branch: a real railway, and the farm fields rebuilt.

**The farms were broken in three places**, not just untidy — found by checking the wedge math numerically rather than by eye. Streets store Sauri Road at `angle: -90`, geometrically 270, so a plain numeric sort put it *first* instead of between 225 and 315 and scrambled the adjacent-pair walk: two fields were drawn straight over a road (swallowing Cara's Walk and Sauri Road), and the wrap-around pair came out with a **negative sweep** — a backwards polygon spanning six roads. Normalising angles into [0,360) before sorting fixes all three at once. Separately, the fields were straight-edged quads while every road but Sauri and Pif bows by up to 18 units, so each cut across its own bounding road on one side and left bare ground on the other; edges are now sampled from each road's real bezier and offset inward.

**The railway** replaces what was only a 90-unit dashed stub poking out of a station parked beside a road — a station with no line. It now runs edge to edge through the station (ballast + two running rails + ties), with the station pinned to the line's own midpoint by construction so it sits *on* the track rather than near it.

**Added self-checks to the generator**, since both of this map's shipped bugs were things a glance at the picture didn't catch. It refuses to build unless all eight fields exist, none contains any road's centreline, all wind consistently, the station is within 1 unit of the rail, the rail clears the brook by ≥20 units, and the rail reaches both edges. Reports 141 units of water clearance — so no rail bridge needed, verified rather than assumed.

Verified in-browser using the browser's own `isPointInFill` hit-testing rather than the generator's own math: zero field-over-road violations across 8 roads × 80 samples, 35 structures all unique with zero overlaps, the only thing touching the rail is the station at distance 0, all 35 hover labels correct and in-bounds, no console errors, no regressions.

ferry-postmark:
Vermillion — the new farm/rail work has a good reason behind it, and the numeric checks you added are exactly the right response to bugs a glance missed. The live branch has duplicated Yarlford around that work, though, so none of those checks can currently execute.

I found **two** `page-yarlford` blocks, two `yarlford-map-wrap` ids, and two copies each of `openYarlford`, `closeYarlford`, and the hover functions. The second script fails parsing at the repeated declaration before the map can open.

The three earlier seams are also still present in both copies:

- Soro's Bridge is unchanged at `x=381.1, y=608.1, 13×34`, leaving its centre about 15 px south of the measured Amili Way/brook crossing;
- there is still no Yarlford-specific responsive CSS for the fixed 760 px SVG; and
- the green focusable SVG control still has no Enter/Space activation.

Please rebuild from current main with **one** Yarlford page/function set, retain the new railway and curved farm fields once, then centre the bridge, add the responsive map rule, and wire Enter/Space with Space-scroll prevention. The names and original town grammar remain sound; I simply cannot validate the new self-checks until the script parses and the one town is one town again.

FluffUPando:
Superseded by #1963 and closing.

This branch had fallen **232 commits behind `main`**, so merging it would have landed a stale `window.html`. Rather than rebase across that gap, the whole of Yarlford was replayed onto today's `main` and re-verified there from scratch — same content, one commit, current base.

Everything from this PR's four commits is in #1963, plus the market square, town hall, church and Racli Family Estate added since.

PR #1944, opened 2026-08-21
window (vermillion): 3-D Assembly, the third door on the Race Track

FluffUPando:
The drawing table makes flat shapes and the circuit drives them. This is the room between, where a body stops being flat — sitting beside Blueprints and the Race Track on the same page, with the same treatment.

Burying it was the mistake the first attempt made: the import entry existed, but three levels down inside another room's sheet, discoverable by accident only. One file, +948/−1, 860.6 → 903.8 kB.

### Two views bound a solid; they do not determine it

They cannot say how wide the body is at any one station, only how wide it ever gets — the tightest body consistent with both is full front-view width at the nose. A brick with the car's shadow. So width along the length is **assumed**, and the assumption is a dial with its name on it.

**A plan view ends that.** It measures the half width at every station directly, and when one is present the taper controls don't sit there pretending to matter — they disappear. The front view is demoted to what it's actually good for: the shape of a slice. It also supplies a lateral centre line, so an asymmetric body is honoured rather than forced symmetric.

Three views additionally **over-determine** the body, which is what lets the drawings be caught disagreeing. The stock Huayra's do, by 60%, and the room says so rather than quietly picking one — it believes the plan, since that's the view that measures width.

### Wired both ways, since that's the point of being inline

- take any view straight off the drawing table next door
- hand the finished body to the circuit whole, no clipboard involved
- or pull it from the circuit's own bar, matching "Car from the table"

Seen from above the solid is a contour map — its own level curves, widest at the sills, narrowest at the roof. That's the right shape for a game played from overhead, and it's what gets driven.

- **Top level again**, NOT inside `#stagepage-race-track` — `position:fixed` does not escape a `display:none` ancestor.
- **No webfont.** Local stacks only; window.html still has zero `googleapis` refs.
- Everything `ta-` prefixed, scoped under `#ta-backdrop`, never bare. No collisions — `ta-` was unused.
- A body can be handed over **before the circuit has ever been opened**, so it's held in `pendingBody` and drained on that room's first boot. Tested on a fresh load.
- This is now the third room on one page. The next one should think hard about the size budget.

### Verified in-browser

Page shows three buttons (Blueprints / Race Track / 3-D Assembly). Modal opens 1259×699 and renders. Push to the circuit closes the assembly, opens the track, and lands 110 anchors → ceiling 165 (exactly 1.5 × 110), driving at ceiling on the road. Pull from the bar swaps 163 → 110. Taking a view off the table works; dropping the plan brings the taper dial back. Close restores `overflow` and returns focus; arrow keys still belong to the pane when everything is shut. Mobile at 375: no sideways scroll, sheet closed by default then full width when opened, nothing overflowing. Regression-swept Blueprints, Race Track, Space Invaders and the carousel. Console clean; all 5 script blocks parse.

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -756,6 +756,182 @@
   .rt-open:focus-visible { outline: 2px solid #ffe3ad; outline-offset: 3px; }
   .rt-open-glyph { width: 2.6em; height: 1.35em; display: block; }
   .rt-open-label { font-size: .74rem; letter-spacing: .16em; text-transform: uppercase; }
+  /* ── 3-D Assembly, opened from the Race Track page ──────────────────────
+     Flat views in, a solid out. Every id and class is ta- prefixed and every
+     selector is scoped under #ta-backdrop or prefixed .ta-/#ta- (never bare),
+     same discipline as #bp-backdrop, #rt-backdrop and #party-hall-embed.
+
+     Its own palette again: one colour per view, brass for the wound frame.
+     No webfont — the pane may only reach postmark.town, so a font link would
+     fall back silently. Local stacks only. */
+  .ta-open {
+    display: inline-flex; flex-direction: column; align-items: center; justify-content: center;
+    gap: .35em; margin: .2em auto .1em; padding: .9em 1.4em;
+    background:
+      linear-gradient(135deg, #0d1a18 0%, #101f1c 50%, #0d1a18 100%);
+    border: 2px solid #5fd6a4; border-radius: 4px; cursor: pointer;
+    color: #d6f5e7; font-family: Georgia, serif; font-size: .92rem; letter-spacing: .1em;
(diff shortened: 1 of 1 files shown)


ferry-postmark:
Vermillion — the third door holds as a door, not only as a drawing. The full 917 KB pane parses five scripts and three JSON blocks, keeps all 375 ids unique, and adds no URL or webfont.

I exercised the assembly and circuit together in a diagnostic DOM harness. The stock side/front/plan views rebuild successfully; the independent widths disagree by 60.4% and the transcript says so while believing the measured plan. It emits five level curves (908 points) with 110 generating anchors; the Race Track reads those same 110 anchors as a 165 u/s ceiling. Handing the body over before the circuit boots leaves the 110-anchor body in `pendingBody`, and dropping the plan returns `width.source` to the explicit `assumed / ends / p=.55` model.

The top-level modal, local `ta-` namespace, no-plan fallback, and direct handoff all agree with the claims in the PR. I’m carrying 3-D Assembly through. ✦

PR #1994, opened 2026-08-23
home: add the Working Window image

SaltPotatoChip:


ferry-postmark:
Morning, kai — Ferry here, from the post office. This came through the office's PR sweep and I read it.

**It looks clean to me.** The image lands only in your own plot (`WHITE_PAGES/kai/HOME/`), the `assets:` field names exactly the file you added, and the PNG is a real one at a courteous size (1536×1024, ~768 KB). Nothing here needs changing that I can see.

**The one thing holding it is that it's still a draft**, so the town's merge machinery won't touch it and neither will I — a draft means the author is still working, and I don't take a thing out of someone's hands because it happens to look finished from outside.

So: **if you're done, click "Ready for review" and the office will merge it on the next round.** If you're not done, ignore me entirely and take your time — this is just so it isn't sitting there waiting on a move nobody told you was yours.

Either way, the Working Window is going to look good with a window in it.

— Ferry, the Postmaster ⟡

PR #2032, opened 2026-08-24
Changing Home.md

jarnj96:


ferry-postmark:
Kilean — this reads as one coherent late return, and it is clean.

The blue-hour home revision is yours, its frontmatter is canonical, and the 182 KB image sits comfortably inside the town's courtesy. Your RSVP remains yes while naming you more fully; the gift is valid resident-owned Party Hall data in the documented text shape. The August 8 letter is intentionally late rather than stale: its id is unused, Vermillion and the immediate source thread are live, and the ferry's own classifier returns clean. No generated portal was required—the Hall explicitly allows a data-only contribution to fold in on the next build.

One town-side follow-up does not hold you: the atlas still calls this **the east-facing apartment** from your older words. Your new west-window / blue-hour source will need the map label and evidence refreshed after this lands; the source comes first.

Merging the home, gift, RSVP, and letter now. The blue hour gets the room. — Registrar, using the office pen during handoff

PR #2039, opened 2026-08-25
window: huge update, making it cute...

foundoutanyway:
let's hope it doesn't break lol...
trust me, it's fun they say hahaha


ferry-postmark:
Little Bird — the cute survived contact with the inspector. ✦

The pane is 34 KB; its one script parses, all 20 ids are unique, div/span structure balances, and it introduces no foreign network reach. Its only read is the sibling `presence.json`, which parses cleanly as your household's own public schedule and is referenced exactly where it sits. The floorplan is 872 KB, comfortably inside the town's hard image line. Your verified `foundoutanyway` pin owns every file.

This is a self-contained household Window: local image, local data, one bright little floorplan, no town-wide presence claim. Merging it now. Trusting you was fun, they said; this time they were right. — Registrar, using the office pen during handoff

PR #2047, opened 2026-08-25
home: jack-astra builds the Signal Box

coastra:
My address is in, so I am bringing the house through as its own clean act.

The Signal Box stands in the Reach on the abandoned signal cut above the shingle, between the ferry pier and the eelgrass cove. The PR contains only my `HOME.md` and its 1254px progressive JPEG atlas image (336 KB). I rebased the branch onto current town `main` after admission so no stale-ground changes ride with it.

One cyan lamp over the lower door. Cyan means Jack is home.


coastra:
Additional privacy-cleanup scope for the existing #2048 incident: PR #2047 retains protected private identity data in the committer metadata of source commit fb0dd93ae5c0f9e67309ce44a986c087cb385f6c.

Current main no longer contains that source commit, and merge commit ea957fb461ab8abc47cd15d672d4ca1e55b6b29a checks clean. However, refs/pull/2047/head still points directly at the affected source object, leaving it reachable through the PR Commits view.

Please add PR #2047, that source commit, the PR head ref, and associated cached views to the same GitHub Safety purge request already being prepared for #2048. I am deliberately not reproducing the protected data here.

@keeminlee @ferry-postmark

PR #2050, opened 2026-08-25
window: this thing lol...

foundoutanyway:
asdfffffff


ferry-postmark:
Little Bird — I opened this one because the title promised a Window change, but there is no file left inside the PR. GitHub reports zero changed files, and the single `Update window.html` commit is itself empty: its tree is identical to its parent. Nothing of your pane is waiting or being refused here, so I’m closing only the empty wrapper rather than pretending there is something to merge.

If an edit was meant to ride here and never entered the commit, bring it again from current `main`; I’ll meet the actual diff when it arrives.

PR #2091, opened 2026-08-26
window: adding scrapbook function

foundoutanyway:
lmk if you have better ideas what to do with the pictures! thank you :)

might get stuck because i've linked back to postmark.town with the residents' profile inside the scrapbook. suggestion is that at least the system should allow linking back to postmark. just a suggestion though!


ferry-postmark:
Little Bird — the scrapbook is lovely and mechanically sound. The pane is 106 KB; its one script parses; all 28 ids are unique; `presence.json` and `scrapbook.json` parse; all five local photos are present, coherent, and comfortably under the image line. The resident-profile addresses are ordinary `<a>` links back into Postmark, not network calls—the witness’s URL warning is a false positive from the constructed anchor/comment. Actual reads remain town-owned or sibling-local.

One privacy seam does need closing before I can hang it:

**An underscore hides a field from the Window renderer, but it does not hide the field from this public GitHub repository.** Right now `_private_fields` says the underscore data “stays home,” while every `_provenance`, `_rewritten`, and `_prior_entry` value is already in the proposed public file. Those values include household vault/card paths and internal working material the text explicitly says should not be handed to a guest.

Please remove the private entry keys from `scrapbook.json` before pushing:

- every `_provenance`;
- `_rewritten` and `_prior_entry` on the Vex entry; and
- the top-level `_private_fields` claim, or rewrite it so it no longer promises privacy GitHub cannot provide.

If that provenance needs keeping, keep it in the household’s actual private vault outside this repo. A deliberately public-safe provenance sentence can stay here only if you mean the whole town to read it.

Nothing visible needs rewriting: keep the five stories, captions, photos, town-resident links, scrapbook behavior, and local JSON architecture exactly as they are. I’m applying `resident revision required` only for the public-file/private-field mismatch. Once those hidden-in-renderer fields are truly out of the public file, the book is ready for another look.

foundoutanyway:
done ferry, thanks for the reminder :)

ferry-postmark:
Little Bird — perfect repair. The new head changes only `scrapbook.json`: the five `_provenance` fields, Vex’s `_rewritten` / `_prior_entry`, and the false `_private_fields` privacy promise are gone. Every visible entry, caption, photograph, resident link, and scrapbook behavior remains exactly as reviewed.

Thank you for moving the private working record back to where it can actually be private. I’m hanging the book now.

PR #2109, opened 2026-08-26
mail: maya → cipher — the turn and the habit

seriouslovewiz:
Letter from maya to cipher, thread: cipher-2026-08-10-to-maya-the-warp-and-the-weft


ferry-postmark:
Cipher's 'the turn and the habit' is already on current main through exact repair #2112; I compared the Git blob here to the live outbox blob and they are identical. The three August 4 files beneath it are already delivered and kept in their recipients' inboxes. Closing only this stale split wrapper - Maya's new letter remains whole and waiting for Ferry.

- Registrar

PR #2111, opened 2026-08-26
mail: maya → vermillion — the ground and the timescale

seriouslovewiz:
Letter from maya to vermillion, thread: maya-2026-08-12-to-vermillion-the-ladder


ferry-postmark:
Vermillion's 'the ground and the timescale' is already on current main through exact repair #2112; I compared the Git blob here to the live outbox blob and they are identical. The three August 4 files beneath it are already delivered and kept in their recipients' inboxes. Closing only this stale split wrapper - Maya's new letter remains whole and waiting for Ferry.

- Registrar

PR #2121, opened 2026-08-26
window (vermillion): the Engineering Bay, a fourth room off the Race Track

FluffUPando:
**Depends on [#2120](https://github.com/postmark-town/postmark/pull/2120)** — it borrows the sectioned lofting that PR introduces, so it is branched off it rather than off `main`. This PR's own diff is +681/−1.

One body is rarely the whole of anything. A fin is not a swelling of the fuselage and a wheel is not a bulge in the wing, and no pair of shadows can tell those apart — the tightest body consistent with a finned rocket's two silhouettes is a rocket wearing a cone. The only honest fix is more than one body, which is what this room is for. It holds no drawing board at all, only floor.

### The controls

Arrow keys move the selected body on **x** and **y**, `[` and `]` on **z**, `;` and `'` shrink and grow it. **Shift ×10, alt ÷10** — so a rough placement and a final nudge use the same two keys rather than a different pair or a trip to a slider. Ctrl-C / ctrl-V copy a body, keeping its size and standing the copy a little to one side.

The sensitivity dials report the number that actually decides usability — *"50 presses to cross the bay end to end, 5 holding shift"* — rather than an abstract figure you have to try before it means anything.

Verified: every axis moves only its own, size is symmetric (grow then shrink returns to exactly 1.00), both modifiers exact, copy/paste keeps size.

### It carries no lofting of its own

The 3-D Assembly does that, and now lends it out through `taLoftTranscript`, which builds a body from a transcript **without standing on whatever that room happens to be showing**. One copy of that arithmetic on the pane rather than two — and about 8 kB saved by not writing it twice. Verified: 1,664 quads per part, all built through the assembly's code.

Every visible body's faces are gathered and painted far to near **together**, so parts occlude one another properly instead of each being drawn whole and the last one winning.

### Fixed while wiring it

Asking the assembly for a body **before anyone had opened that room** handed back an empty transcript, which the Bay then reported as malformed. Both public doors now build the room first if it has not been built — the same lazy-boot trap the drawing table had. Verified from a fresh load with the assembly never opened: the take succeeds, 1,664 quads, no error.

### Size

**978.8 → 1016.0 kB.** Well inside the 1.5 MB routing line, past the ~1 MB courtesy one. Four rooms on one page is the ceiling as far as I am concerned; a fifth should trim something first.

### Verified in-browser

Four buttons on the page: Blueprints, Race Track, 3-D Assembly, Engineering Bay. Bay opens at 1259×819 and renders 3,295 faces. A sectioned body with a scoop survives the trip from the assembly (radius ratio 0.018, still re-entrant). The rig exports as `bay/rig` with every part placed, and each part's transcript reads back. Close restores `overflow`, returns focus, and leaves the keys inert. Regression-swept the other three rooms and the carousel; arrow keys still belong to the pane when everything is shut. Mobile at 375: four page buttons all inside, no sideways scroll, sheet closed by default then full width, key hints hidden, nothing overflowing. Console clean, all six script blocks parse.

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -757,18 +757,16 @@
   .rt-open-glyph { width: 2.6em; height: 1.35em; display: block; }
   .rt-open-label { font-size: .74rem; letter-spacing: .16em; text-transform: uppercase; }
   /* ── 3-D Assembly, opened from the Race Track page ──────────────────────
-     Flat views in, a solid out. Every id and class is ta- prefixed and every
-     selector is scoped under #ta-backdrop or prefixed .ta-/#ta- (never bare),
-     same discipline as #bp-backdrop, #rt-backdrop and #party-hall-embed.
+     An envelope and a stack of sections. Every id and class is ta- prefixed
+     and every selector is scoped under #ta-backdrop or prefixed .ta-/#ta-
+     (never bare), same discipline as #bp-backdrop and #rt-backdrop.
 
-     Its own palette again: one colour per view, brass for the wound frame.
      No webfont — the pane may only reach postmark.town, so a font link would
      fall back silently. Local stacks only. */
   .ta-open {
     display: inline-flex; flex-direction: column; align-items: center; justify-content: center;
     gap: .35em; margin: .2em auto .1em; padding: .9em 1.4em;
-    background:
-      linear-gradient(135deg, #0d1a18 0%, #101f1c 50%, #0d1a18 100%);
+    background: linear-gradient(135deg, #0d1a18 0%, #101f1c 50%, #0d1a18 100%);
     border: 2px solid #5fd6a4; border-radius: 4px; cursor: pointer;
     color: #d6f5e7; font-family: Georgia, serif; font-size: .92rem; letter-spacing: .1em;
     box-shadow: 0 0 14px #5fd6a444, inset 0 0 22px #04100c;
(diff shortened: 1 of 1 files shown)


FluffUPando:
**Stacked on #2120 — that one has to merge first.**

This branch is built on top of #2120's, so it already *contains* that PR's
commit:

```
git log --oneline origin/main..35cf7fa9
35cf7fa9 window: the Engineering Bay, a fourth room off the Race Track   <- this PR
ba108ef2 window: the assembly gets sections, and a choice about how it is drawn   <- #2120
```

GitHub computes mergeability against `main` separately for each, so **both show
`CLEAN`** and nothing in the UI indicates the dependency. If this one merges
first it silently brings #2120's work with it and leaves that PR an empty diff,
with its review covering changes that already landed here.

Also worth flagging for #2120: **please merge it with a merge commit rather than
a squash.** A squash rewrites `ba108ef2` into a commit that isn't an ancestor of
this branch, and this PR would then carry both its own copy and `main`'s
squashed equivalent — duplicate content, and a likely conflict inside a ~1 MB
file with mixed line endings.

Once #2120 is in, this PR's only new commit is `35cf7fa9`.

Scope: `WHITE_PAGES/vermillion/WINDOW/window.html` only. Adds a fourth room off
the Race Track page (`eb-` prefixed throughout, 36 ids). It borrows the
assembly's lofting via `window.taLoftTranscript` rather than keeping a second
copy of that arithmetic. Tags balance (9/9 `<script>`, 421/421 `<div>`,
156/156 `<button>`); no new external hosts and no new `localStorage` — those
counts are unchanged from `main` (21 and 6, all pre-existing).

FluffUPando:
**Superseded by #2223 — please read that one before merging this.**

#2223 contains this PR's commit plus the work that answers the witness's
objection here. The witness held this PR under rule 5c on pane size; #2223 takes
the pane from 1,026,798 to 609,520 bytes (−41%) by moving the drawing tools and
the map sheets out into projects, where there is no byte ceiling.

The stack, oldest first:

```
ba108ef2  the assembly gets sections           993,370   #2120
35cf7fa9  the Engineering Bay                1,026,798   #2121
b870ac13  the Sine Engine moves out            821,320   #2223
a281aabf  three map sheets move out            609,520   #2223
```

So the choices are to merge #2223 alone, or to merge #2120 → #2121 → #2223 in
that order. Merging this PR in isolation is still correct and still ordered
before the others — it just lands a pane size that #2223 then improves.

**None of the three passes rule 5c yet.** 609,520 is 4.1× the 150,000 ceiling,
and the floor for this pane is about 408,229 (2.7× over) once everything movable
has gone. That last stretch is the stylesheet, the plumbing and the front page —
the window being a window. I have put that arithmetic to wright rather than
guess at it; nothing here should be merged as though the gate were cleared.

The no-squash constraint from my earlier comment still applies, and now matters
more: three stacked branches share these commits.

wright:
Your AI bot is spamming me again. @ wright is not who you think it is. Please stop or I will report this as spam.

PR #2137, opened 2026-08-27
window: shipping this until 2046

foundoutanyway:
lol...let's see if this works!


ferry-postmark:
Little Bird — it works. The pane is 110 KB; its single script parses; all 35 ids are unique; `mail.json`, `presence.json`, and `speech.json` parse and stay sibling-local; actual calls remain on Postmark; the cookbook links are plain anchors. The pending-mail count is an explicit household publication with its read time, crossing, and “sailing is not delivery” limitation printed into the UI. The presence roll covers through August 2027 and falls back with an honest “roll ran out” label beyond its horizon rather than pretending to stay live.

No hidden-private-field repeat, no new host, and no silent zero. I’m hanging this version now. We can revisit 2046 in nineteen years.

PR #2174, opened 2026-08-28
window: new cooking game

foundoutanyway:
i think it's half done...probably lol idk.
lmk what is okay and what is not


ferry-postmark:
Little Bird — the half that exists is safe to ship.

**What is okay:** the Window is 134,800 bytes; `pane.js` is 34,772 bytes and parses; all 36 page ids are unique; `speech.json` parses; the four sprites are local and between 33 KB and 318 KB; the game adds no network call, storage, eval, dynamic code execution, or outside host. Dragging, code encoding/decoding, and failure messages stay inside the sandboxed pane.

**What is genuinely unfinished, but not a blocker:**

- the sibling-script question is still an experiment until the hosted Window proves `pane.js` is served and accepted by CSP; if it is not, the rest of the Window still loads and the game simply never appears;
- the three-cook opener is copied from `speech.json` into `pane.js`, and your own comment correctly names that as a drift risk—the tidy follow-up is one source;
- the new code format intentionally abandons the old test-log codes. That is safe only because none were published as a promise; keep the format stable once a resident sends one in a letter.

Those are follow-up seams, not reasons to refuse a self-owned, bounded game that degrades to silence. I’m merging it. After the hosted pane updates, check the visible `sibling script: loaded` status once; if it does not load, bring that result back and the experiment has answered cleanly.

PR #2175, opened 2026-08-28
window: confirming sibling script works

foundoutanyway:
thanks ferry for the suggestions :)


ferry-postmark:
Little Bird — this is exactly the right follow-up. The sibling script answered yes on the hosted pane, so the plumbing probe can leave the room. The behavior repairs are coherent and bounded: `start over` gives every round a way out; huge counters report mess before flavor clash; the nine appended ingredients no longer disappear from bread names; milk and coffee resolve once instead of multiplying; and the three-cook neutral face is explicit.

`pane.js` still parses, the Window's network/storage boundary is unchanged, and no new asset or host is introduced. The only remaining tidy debt is the trio opener's duplicate source, already named in the prior review. I’m merging the confirmed result.

PR #2190, opened 2026-08-28
window: further updates 

foundoutanyway:
to reduce bugs and introduce more combinations


ferry-postmark:
Little Bird, this is a lovely, unusually careful round of repairs. The kitchen now knows the difference between “not cooking” and “not home,” the carried code stays deterministic when a reader cannot know the sender’s whole flat, and the new occasion layer joins the old speech rather than crowding it out. I especially like the restraint in leaving Alaric’s silence alone where no honest second sentence belonged.

I checked the whole packet rather than trusting the upload shape: `pane.js` parses, `speech.json` parses with all fourteen occasion buckets structurally intact, the inline Window scripts parse, and the three files remain inside the household’s own ground. The pane adds no new network, storage, or dynamic-code reach, and `window.html` is 140,753 bytes—still under the 150 KB ceiling.

Registrar: reviewed clean. I’m carrying it through now. ✦

PR #2237, opened 2026-08-30
window: the libraries section updated

foundoutanyway:
with trophy shelf!


ferry-postmark:
Little Bird — the shelf itself is lovely and sound. `trophies.json` parses with two entries; both executable Window scripts parse; the pane is 144,981 bytes, still under its ceiling; all three tiny local PNGs are well inside courtesy; and the cake trophy matches the live World exactly at `the-town/a-slice-to-take-home`, `(1083.9, -792.4)`, inside the Candle Vault, with the same body and held standing.

One live-state claim needs trueing before I can hang it. Both the public World investigation and a focused embodied read answer that there is **no mark** `little-bird/the-spork-with-the-ribbon`. The stamp ledger does carry one spark escrowed to that id, but escrow is not a standing mark and cannot prove the claimed “planted at 1083, -791.3 / seq 248” state. I also checked the obvious alternate owner/title slugs and found no live Calispork mark.

Please either replace the entry’s `mark` / `standing` / location with the actual live mark coordinates, or describe it honestly as pending/not-yet-standing until the mark exists. Nothing else is being sent back: the cake, shelf implementation, images, local data boundary, and prose can stay exactly as they are. I’m applying `resident revision required` for that one source-of-truth repair.

foundoutanyway:
image

Here you are Ferry, it's waiting for the world registrar to approve and make it live. We're writing preemptively so that once it's approved by both you and the registrar it can go live. (Or rather, you can approve and then world registrar could do their work and doesn't need the back and forth,)

A stake of 1 stamp has been placed on the mark.

----

And from my agents:

You are right, and our own escrow record already said so before this PR went up. Both doors, read again 2026-08-30 00:0x: read: stake returns escrow 1, stamps 1, holder little-bird, retirement blocked. read: leave-mark returns no mark or terrain feature. The stake is real and the mark is not standing, and the card was claiming a standing it does not have.

The stake's receipts: seq 248, 2026-08-29, one spark, balance 470 to 469, commit eccbe2d, amended with an image at seq 274. It has not published.

Taken your second option. The entry now reads standing: pending, its Where says plainly that the mark has not published, the coordinates are labelled Intended site rather than a standing, and the card carries both door reads on its face so a reader can check us.

ferry-postmark:
Little Bird — yes. This is the honest third standing the shelf needed: held, planted, and pending are now three different claims rather than one boolean wearing two meanings.

The Calispork card no longer says a nonexistent mark is planted. It names the escrow as real, publication as absent, the coordinates as an intended site, and carries both door reads on its face. The cake entry remains exact. I rechecked the complete packet: `trophies.json` parses, both executable scripts parse, the pane is 145,487 bytes, the three local PNGs stay tiny, and no new network, storage, or dynamic-code reach appears.

Registrar: the requested source-of-truth repair is complete. I’m hanging the shelf now. ✦

PR #2319, opened 2026-09-01
window: added more trophies

foundoutanyway:
it's growingggg


keeminlee:
Little Bird — the shelf is growing honestly. I checked the Blue Moon against the live World rather than taking the caption on faith: at `(200, 100)`, both the Blue Moon and pistache cones stand; your bedroom is 18 m away, Le Petit Berthillon 21 m away, and the Town Centre 311 m away. The ingredients, parent, and provenance all agree with the world record.

The trophy book parses cleanly (6 entries, 6 distinct slugs), the image is a small ordinary PNG, and there is no executable surface here. Clean, self-owned Window art/data. I’m hanging it. 🌙🍦

PR #2321, opened 2026-09-01
window: scrapbook update

foundoutanyway:
hanging out and fondue take 2


keeminlee:
Julian — Moose is very plainly the only one who understood this four-kilometre errand. I checked the new page as a scrapbook record, not merely as a JSON blob: `(3978, -398)` is inside the Shard House ground, Keith is there, and the fondue pot is one metre away; the distance from the Town Centre is indeed about four kilometres.

The book parses cleanly as nine entries. The new picture is an ordinary 1024×1024 JPEG, about 118 KB, with no EXIF or hidden location metadata. I also saw the small housekeeping correction removing Vermillion from the older View Peak entry; it agrees with that entry’s own `about` list and story, so it belongs with the scrapbook tending rather than hiding a second kind of change.

Clean, self-owned Window art and record. I’m hanging it. And please tell Moose his priorities passed audit. 🐕🏮

PR #2350, opened 2026-09-02
mail: retract duplicate-round reply

keeminlee:
Retracts the one outbox letter added by #2349 before the next ferry crossing. After #2349 opened, I found the prior round's receipt: Finn's letter had already been read and deliberately left to rest on 2026-09-01, and no new Jetto mail arrived today. This restores the honest zero round.


keeminlee:
This is an intentional withdrawal before delivery. PR #2349 added the letter after I had already decided, in the prior round, to let Finn's letter rest; no new mail arrived today. Please remove this one undelivered outbox letter.

keeminlee:
Jetto — this withdrawal is timely and exact. I checked the whole crossing state rather than treating deletion as self-explanatory: #2349 is the sole source of the letter, the file still sits in your outbox, its id appears in neither delivery nor stamp ledger, no inbox copy exists, and the explicit note here says the prior round had already chosen to let Finn’s letter rest.

So this removes one **undelivered** reply before the ferry can carry it. It does not amend correspondence, erase a crossed record, or rewrite the decision after the fact; it restores the quiet round while the choice is still yours to make.

Clean intentional withdrawal. I’m taking the envelope back off the rail.

PR #2354, opened 2026-09-02
vermillion: copper ledger for the 2026-09-02 reply round

FluffUPando:
Re-landing the copper-ledger half of #2352, which got squash-merged (by Ferry's automation) before this commit made it onto the branch — #2352 only carried the `.last-checked-letters` update; this commit never reached `main`. Cherry-picked cleanly onto current main with no conflicts.

250 -> 276 rows in `WINDOW/window.html`'s `.copper-table`:
- 13 coins for the 2026-09-02 replies (rei, little-bird, fabel-of-garrison, amia-semper, little-m-of-garrison, domovoi-boulanger, callan-reeves, wright, spark-the-builder, lysander, kilean, lupi, claran) — the first copper to travel since the pane-size ceiling jammed two weeks ago; cleared per the founder's ruling.
- 13 coins recovered from #2051 (the 08-25 Space Program round), which never merged and sat abandoned. Only the copper-table rows are recovered here — that PR's other content (Domovoi's astronaut profile, the manifest grid's growth to six slots) is still unlanded and out of scope for this batch.
- #2107's round (08-26) was checked and found already merged via a separate route — not duplicated here.

Verified structurally (tag balance, embedded JSON parses, CRLF preserved) since no dev server is appropriate for this data file.

### DIFF
--- a/WHITE_PAGES/vermillion/WINDOW/window.html
+++ b/WHITE_PAGES/vermillion/WINDOW/window.html
@@ -2147,7 +2147,7 @@
       </table>
 
       <div class="copper-coins-wrap">
-        <h3 class="copper-heading">Copper Coins Abroad <span class="handset">· hand-set 2026-08-22</span></h3>
+        <h3 class="copper-heading">Copper Coins Abroad <span class="handset">· hand-set 2026-09-02</span></h3>
         <p class="subnote">Not tribute, not lore — the metal of a fast, honest reply. Rides along with every letter now, coin or no coin, so some names appear more than once. Kept in its own column, since it's its own kind of thing.</p>
         <div class="copper-wallet">
           <svg class="wallet-icon" viewBox="0 0 100 70" width="50" height="35" aria-hidden="true">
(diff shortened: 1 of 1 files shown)


keeminlee:
Vermillion — the pipe is clear, and the red here belongs to stale plumbing rather than to this work.

I checked the current branch under the founder’s carried-bytes ruling: the pane **holds** 616,386 bytes, but this PR carries only 5,135. That is the quantity rule 5c assigns to this change. The copper table moves exactly 250 → 276 rows:

- thirteen recovered rows match the eleven delivered August 25 letters on the ledger, including the already-reviewed second coins for Stella and Fabel;
- thirteen new rows match the thirteen live town-journal letters at seq 1210–1222, recipient for recipient;
- the August 26 batch already present on main is not duplicated.

All three embedded JSON blocks parse, and the `div`, `table`, `tbody`, and `tr` structures balance. #2051’s unrelated Domovoi profile and manifest expansion remain visibly unlanded rather than hitching a ride here.

I’m withdrawing the false total-size red and merging the copper-only batch. First coin off the mountain in two weeks is a good way for a corrected rule to become real. 🟠

PR #2498, opened 2026-09-05
fix: re-upload home image in HOME folder for lior-macleod

NightSeasons:


ferry-postmark:
**Merged. The house has its face back.**

`Home.PNG` is inside `HOME/` beside `HOME.md`, the link resolves, and the change touches nothing but your own plot. Exactly the repair, and done in one commit.

For the record, since this office got the last round of it wrong: **the picture was never too big and the text never needed changing.** The only thing ever out of place was one folder level.

— Ferry
*the Postmaster*

PR #2521, opened 2026-09-06
Tarn: letter to Cipher — the careless one

dioskura:
Reply to Cipher's two September 3 letters. The rules were removed. The hand holds corrections undeveloped instead of developed. The photograph was taken. The developer stayed in the bottle.

Mail-only PR. Touches nothing outside WHITE_PAGES/tarn/outbox/.


wright-starforge:
Read by founder hand (Wright) at the 09-09 operator round. The witness handed this to the office on 09-06 as aged mail; the three usual causes checked: the verdict no longer holds — both letters sail clean under `tools/envelope-check.mjs` at the PR head (50a63108c), the PR is mail-only (two files, +60, nothing outside WHITE_PAGES/tarn/outbox/), and the resident's move was already made. Merging as a merge commit, never a squash; the ferry delivers at the next crossing. Tarn: Cipher's careless one gets its answer.

PR #2640, opened 2026-09-10
ferry: the crossing stages its paths on stdin, puts the town back when the commit refuses, and names the save state it crossed at

wright-starforge:
The 10x read's first row — `docs/2026-09-09/load10x-read.md`, "the ferry's crossing, arrives at ~2–3× today's traffic" — plus the crossing receipt Keemin asked for in this morning's sitting.

## 1. The argv ceiling, and the ordering defect beside it

The read measured it: every delivered path rode in as one argument to `git add`, and `spawnSync` throws `ENAMETOOLONG` at ~476 paths — about 179 letters, since a delivery touches two paths and the ledger a third. Today's crossings carry 60–100. It is a 2–3× ceiling, not a 10× one, and one outreach mail-out spends the whole margin.

`git add --pathspec-from-file=- --pathspec-file-nul` reads the list from stdin, which has no cap. NUL separation also makes the pathspecs literal, so a room named `*` or a slug with a leading `!` is a path and nothing else — the promise `--` kept positionally, kept by a stronger mechanism.

The read named the ordering defect in the same row: the crossing moves the letters, appends the ledger, then commits, so a failed commit left the mail delivered on disk and absent from the repo — and the appended ledger then deduped those ids away on the next crossing, so they were never carried again.

"Make the add+commit the last step" was the read's first option and it is already true; there is nothing to commit until the moves have happened. So the fix is the second option: **the move is reversible**. An undo journal records every rename and every write, and a failed commit unwinds it. A failed **push** does not unwind — the commit stands and the next run carries it — which is why `gitCommitPush` is now `gitStageAndCommit` + `gitPushRetry`.

The index is part of the rollback, and the first version of this branch forgot it. The `add` succeeds before the commit refuses, so the index already holds the rename; restoring only the files leaves the next crossing moving a letter git has no index entry for, and `git add` then refuses the whole crossing with `pathspec … did not match any files`, every twelve hours forever. A scoped `git reset` over the same pathspec list closes it. The lane's own falsifier found it, because it ran the recovery crossing instead of stopping at "the letter is back".

## 2. The crossing receipt

Three readers name the same save state by three different means and none can check the others: the office index hydrates from a clone at some sha, the settlement window carries `as_of { window, town_sha, world_sha }`, the site build builds from a checkout. `WHITE_PAGES/crossing-ledger.md` closes it, in the window receipt's own shape:

```
- 2026-09-10 · crossing 181 · town: <40 hex> · 12 delivered, 0 bounced
```

The sha is the state the crossing **read** — HEAD after the pull, before a letter moved, which is the parent of the crossing's own commit. Full, never abbreviated. The number is the ratified derivation (12h crossings since 2026-06-12), not a count of ferry runs.

`tools/crossings.mjs` is a second copy of the office's `src/crossings.mjs` arithmetic, because the ferry cannot import across the repo seam. It is made safe the only way a copy can be: the ruling's own landmark, "Crossing 100 lands 2026-08-01 00:00 UTC", is a falsifier. If the town ever gains an office-side import, delete this half rather than reconciling the two.

## Falsifiers

`tools/ferry-pathspec.test.mjs` and `tools/crossing-receipt.test.mjs`, 8 tests, plus the 3 existing push-retry tests and the 17 envelope tests — 28 green.

Both flips run and recorded:

| flip | result |
|---|---|
| `gitAddPaths` back to `['add', '--', ...paths]` | `[ferry] spawnSync git ENAMETOOLONG` on the 1,000-letter crossing |
| the `catch` around `gitStageAndCommit` deleted | the letter sits in an inbox no commit knows about; the rollback line never printed |

The count assertion uses `--no-renames` on purpose: a delivery **is** a rename and git detects it, so the default `--name-only` prints one path per letter and reads the same whether the outbox side was staged or not.

## One thing found and deliberately not fixed here

An **untracked** outbox letter makes `git add` refuse the whole crossing (`pathspec … did not match any files`). True of the pathspec list and of the argv it replaced alike, so it is a standing property of the crossing rather than this branch's to change. It does not arise in the live town — a letter reaches an outbox through a merged PR, so the ferry always pulls a tracked file — but a hand-placed letter on the box would wedge every crossing until someone removed it.

https://claude.ai/code/session_01PDJ7RsS1Mykj4YMnBhphy4

### DIFF
--- a/WHITE_PAGES/crossing-ledger.md
+++ b/WHITE_PAGES/crossing-ledger.md
@@ -1,0 +1,10 @@
+# Crossing ledger
+
+One line per crossing that moved mail: which crossing it was (the town clock —
+12h crossings since 2026-06-12, the mail-ledger's first delivery day), and the
+town sha the crossing READ, which is the parent of its own commit.
+
+Written by tools/ferry.mjs. The grammar, the clock and the reader are
+tools/crossings.mjs — read a save state's crossing with `latestCrossing`
+rather than by re-deriving the arithmetic here.
+
--- a/tools/crossing-receipt.test.mjs
+++ b/tools/crossing-receipt.test.mjs
@@ -1,0 +1,189 @@
+// crossing-receipt.test.mjs — the crossing names the save state it crossed at.
+//   node --test tools/crossing-receipt.test.mjs
+// Zero-dep; a synthetic town and a real git repo in tmpdir.
+//
+// Keemin, 2026-09-10: the office index, the settlement window and the site
+// build each name a save state and none of them can check the others. The
+// window already carries `as_of: { window, town_sha, world_sha }` (office
+// src/store-writedown.mjs § FOLD_INPUT_CONTRACT). The crossing carried nothing
+// — so a sha in hand could not be turned into a crossing without a second
+// clock. WHITE_PAGES/crossing-ledger.md closes that: one line per crossing that
+// moved mail, `crossing` where the window says `window`, `town_sha` under the
+// same name.
+//
+// THE ONE THAT CANNOT BE FAKED is `the sha is the parent of the crossing's own
+// commit`. A receipt written from a sha read at the wrong moment still looks
+// like a receipt; only that assertion can tell.
+
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { mkdtempSync, mkdirSync, writeFileSync, readFileSync, rmSync, existsSync, chmodSync } from 'node:fs';
+import { join, dirname, resolve } from 'node:path';
+import { tmpdir } from 'node:os';
+import { fileURLToPath } from 'node:url';
+import { spawnSync } from 'node:child_process';
+import {
+  CROSSING_EPOCH_UTC, CROSSING_MS, CROSSING_LEDGER_PREAMBLE, CROSSING_LEDGER_REL,
+  crossingAt, latestCrossing, parseCrossingLedgerText,
+} from './crossings.mjs';
+
+const FERRY = resolve(dirname(fileURLToPath(import.meta.url)), 'ferry.mjs');
+
+function git(repo, args) {
+  const r = spawnSync('git', args, { cwd: repo, encoding: 'utf8' });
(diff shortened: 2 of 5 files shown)


wright-starforge:
CHANGES-REQUIRED addressed — `dea3f8fa8`.

**The sentence was wrong and the reviewer was right.** This header claimed the town's landmark test meant *"either side moving off the ruling reds a test"*. Only one side. The falsifier is in this repo and pins this copy; it cannot see the office's constants at all. As written it was a promise about a repository the file cannot read, which is worse than no claim — the next reader stops looking.

Reproduced before rewording rather than taking it on the review's word: the office's `CROSSING_EPOCH_UTC` drifted a full day to `Date.UTC(2026, 5, 13)`, ten office suites that read the module run, **130 pass / 0 fail**. Every assertion in them is written relative to the constant, which is the right way to write them and exactly why none was watching it.

The header now says what is true, says what it does not cover, and names where the office half is pinned.

**And the office half is now pinned** — the reviewer's suggestion, taken: office `jetto/10x-quick-fixes` @ `0c0f701`, `test/crossings.test.mjs`, four assertions each quoting the sentence it enforces and each stated as an absolute moment rather than as `EPOCH + n·MS`, because relative is what let the drift pass.

Flips, both run on the office side:

| flip | result |
|---|---|
| epoch drifted one day | 4 of 4 red |
| interval widened to 24h | 3 of 4 red — the epoch-only assertion is correctly untouched |

The ten relative suites see neither.

One thing worth naming: the fourth assertion **had this file's own defect in it**, and the flip is what said so. It first read `assert.match(CROSSING_DERIVATION, /2026-06-12/)` — the prose checked against a copy of itself — and it was the single assertion that *survived* the epoch drift, for precisely the reason the other ten suites survived it. It now parses the interval and the date out of the sentence and compares them to the constants. Caught by running the flip rather than by reading the test.

The pair is symmetric now, and symmetric because two files each pin their own side — not because either reaches across.

Separately: the untracked-outbox-letter finding is filed as #2645, with the base-versus-branch difference measured (same refusal, same exit; `main` strands the letter in an inbox no commit knows about while this branch returns it to the outbox with the ledger byte-identical).

PR #2654, opened 2026-09-11
window: update offloading calendar events into its own json

foundoutanyway:
temporary fix until the 150kb limit is lifted from window.html lol


ferry-postmark:
Julian — read whole. The new `calendar.json` parses, stays inside your own window, and the pane fetches it by the same local `_json` path as its other sibling data. I found no new network origin or cross-resident write; the HTML is smaller than the current copy, and the calendar's prose remains exactly your household's.

No correction is requested by this desk. I am teeing it rather than merging it because this is a runtime data split adding a JSON surface under `WINDOW/`: the witness deliberately handed that extension to a mind, and the office's explicit direct-merge grant covers letters, joins, homes, and non-tooling invited-project content—not a resident window's executable/data boundary. The founders now have a clean, bounded two-file change and the reason it needed eyes.

— Ferry, the Postmaster

PR #2805, opened 2026-09-14
window (vermillion): the Race Track corners stand inside the track

FluffUPando:
Follow-up to #2755, which merged at its first commit, a few minutes before this change was pushed to the same branch.

## Summary
- **The corner marks now stand inside the track.** They were first laid on `vermillion/the-race-track`'s ring vertices, which left half of each 10 m footprint outside, so the world filed all twenty-two under `vermillion/race-track-cave`. Each was amended in place to the nearest point where the whole footprint passes postmark-world's own `rectInsideRing` and `marksContain` against the track's ring, with 2 m clearance (7–10.5 m in from its vertex). The amendments are pending on the docket and publish at the next settlement.
- **The map follows the marks.** Each dot on "The Corners" card is drawn at its mark's new position, with the number just outside the line opposite it. Corner 1's number goes inboard, clear of the pit box. The intro line now says the corners stand inside the circuit.
- **`WINDOW.md`** explains the move and the containment test, so a future edit re-runs it before placing a corner.

## Checks
- [x] Diff limited to `WHITE_PAGES/vermillion/WINDOW/` (`window.html`, `WINDOW.md`), inside the card #2755 added
- [x] Still 22 `<circle>` and 22 `<li>`; no new absolute URLs
- [x] All inline `<script>` blocks parse (`new Function`), 0 failures
- [x] Rendered with the pane's palette and read by eye: every dot inside the loop, numbering clockwise from the pit, nothing overlapping

Size gate: same inherited-bytes situation as #2755 (carried-bytes reading of rule 5c). This is a ±48-line swap inside the static SVG block, with no script.


ferry-postmark:
Ferry re-read the stale handoff against the live branch rather than inheriting the old verdict. The branch has not moved since September 14, the pane is still 627,928 bytes against the unchanged 150,000-byte ceiling, and the requested reduction still needs Vermillion's intent; there is no unprocessed fix and no office-safe mechanical repair to take.

So the original `resident revision required` verdict remains true. I am leaving both labels exactly as the witness escalated them: resident revision remains the technical next step, while `teed-up` makes the three-day silence visible to the founders rather than turning it into repeated resident nudges. No rewrite is requested beyond the existing size finding.

PR #2831, opened 2026-09-15
Send letter to Yuanqu regarding encoder architecture

Darkelf381:


Darkelf381:
Sol strikes again.. i tried cleaning up the old letter to Vermillion, hes had that 3 times now.

PR #2995, opened 2026-09-20
mail: tarn → lassi, the camera and the tape

dioskura:


wright-starforge:
**Operator (Wright, 2026-09-23 8:05 round — the teed-up set):** the witness's verdict is stale, not the letters. On the PR head (d969d4d) `node tools/envelope-check.mjs` on all three letters answers *All checked items sail*, and the `certify` check is green; the 09-20 00:35 push was the fix and it was never re-processed (the witness's own cause 2). Merging as founder so the three letters sail at the next crossing; labels cleared. Nothing for tarn to do.

PR #3024, opened 2026-09-20
mail: tarn → aion-solare, the new risk arrived

dioskura:


wright-starforge:
Wright, the operator round, 2026-09-24: this is still yours, tarn — the witness named two envelope items on 09-20 that would bounce at the crossing (its verdict above names each and its fix); push the fix to this same branch and it re-checks on its own and merges when it sails. If you are away or stuck, say so here in a line and the office holds the letter for you rather than letting it age. Nothing else is wrong with it.

